PR #6, opened 2015-04-06
Use a dynamic wp-load.php path for require_once()

pryley:
Both the plugin directory and the wp-content directory can be moved around in the installation. The location of `wp-load.php` should not be assumed.


AGME-EnterprizeLLC:
Is it possible to develope in central repository for different sites and
servers?

ultimatemember:
We've changed the core code for patch release soon.

PR #28, opened 2015-11-06
plugin adjustments

purplewhite:
1. Sorting: when using a custom sort criteria such as "role", within each role the members should be sorted by display name.
2. Add CSS class for user role to member profile for further styling.
3. Add CSS class for user role to member div's in in members directory for further styling.
4. Add link to cover picture in the members directory for easier click.
5. Cover picture also in format 1.6:1 needed. 

Please check what you can merge. Thanks!


ultimatemember:
Have you tested these changes against performance? Thanks for the contribution

purplewhite:
No, I didn't expect issues here, since the changes are so small.

Issue #73, opened 2016-01-19
Support for user tags in member directory search


nickburne:
Currently mutli-select fields can be pulled through for search within member directories, but user tag fields (from the user tag extension are not pulling through)

nickburne:
I am looking into doing this and will submit as a pull request.

If there is any advice from anyone on how best to do it, please let me know.

ultimatemember:
This is merged in our github branch now.

PR #74, opened 2016-01-20
Allows setting user tags fields as search filters

shramee:
Allows adding	**Ultimate Member - User Tags** `user_tags` fields as `select` search filters.


shramee:
Addresses #73

PR #85, opened 2016-02-11
adding filter hook to emoji base uri

suifengtec:
adding filter hook to emoji base uri.

people of some countries ( such as China ) can NOT use s.w.org, I think there should be a filter hook.


champsupertramp:
Thanks a lot!

suifengtec:
U R WELCOME!

Issue #95, opened 2016-02-23
Form validation does not respect User Tag custom fields


nickburne:
I have a default registration form with a user tag custom field on it and have set it to required, but it doesn't work.

nickburne:
We're looking into fixing this and will send a pull request.

nickburne:
Same issue on profile forms as well as registration forms.

champsupertramp:
Hi @nickburne , I think this will be fixed in the User Tags extension.

nickburne:
@champsupertramp - didn't know which one it would be solved in (extension, core or both). Happy to have access to user tags github if we can help.

nickburne:
@champsupertramp - looks like we are able to fix this in core

nickburne:
@champsupertramp @jonfalcon hi guys, we have just done a pull request for this. Thanks!

Issue #86, opened 2016-02-11
Option to make profile pic a required part of the profile form


nickburne:
Would be happy to develop this into core (pull request)

Let me know what you think.

Thanks,

Nick

nickburne:
We are about to develop this one. Our approach is as follows:

In Forms > Profile Form > Customize this form, we will add the following option:

![screenshot 2016-02-25 10 51 26](https://cloud.githubusercontent.com/assets/3784875/13317687/4e9bcc46-dbae-11e5-874d-ff75da8257a3.png)

nickburne:
@champsupertramp @jonfalcon hi guys, we have also just done a pull request for this. Thanks!

nickburne:
@champsupertramp @jonfalcon - hey guys - can you let me know if this can go in core, if not I will include it in my own customization plugin. Thanks!

PR #114, opened 2016-03-04
Adressing ultimatemember/ultimatemember#95 and ultimatemember/ultimatemember#86

shramee:


champsupertramp:
I've merged both but i will move the code related to User Tags into its extension.

Thank you for your contribution.

PR #117, opened 2016-03-06
Add support for server query string data after user login

10h30:
Add support for server query string data after user login


champsupertramp:
Hi @10h30,
Can you please provide example URL with your commit?
Why not use `add_query_arg( )` in query string concatenation?

10h30:
Hi @champsupertramp ,
The commit is following the discussion here: https://wordpress.org/support/topic/redirection-after-the-user-has-logged-in?replies=6

I use my Wordpress (with Ultimate Member) as SSO Login for Discourse. With the current UM code, the server query string is stripped down after login, user is not redirected back to Discourse after logging in.

Example: club.yeuchaybo.com

champsupertramp:
@10h30 Thank you. Pushed into master repo.

10h30:
I realized that I added the code at wrong line. Already request another pull request to fix that.

PR #129, opened 2016-04-30
Move the check for um_flush_stop to the flush_rewrite_rules function …

svenl77:
…to avoid add_rewrite_rule issues if other plugins flush_rewrite_rules.

We need to make sure the add_rewrite_rule is running each time otherwise the rewrite roles get lost if other plugins flush_rewrite_rules.

Now the um_flush_stop check is moved to the add_rewrite_rule function to make sure add_rewrite_rule is fired.
If the option "Stop rewriting rules on every load" is set to "on" and a plugin flush_rewrite_rules your add_rewrite_rule will get recognised.

I also added a filer 'up_rewrite_flush_rewrite_rules' to allow plugins to stop rewriting without manipulating the option array.

I hope this makes sense for you and help you. My english is not the best let me know if I got something wrong here or you have any question left. I like to help.

### DIFF
--- a/core/um-rewrite.php
+++ b/core/um-rewrite.php
@@ -32,7 +32,7 @@ class UM_Rewrite {
 
 		global $ultimatemember;
 
-		if ( isset( $ultimatemember->permalinks->core['user'] ) && !um_get_option('um_flush_stop') ) {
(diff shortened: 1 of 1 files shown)


champsupertramp:
Thank you for your contribution.

PR #168, opened 2016-09-10
Make sure $footer_text is returned if( $pagenow !== 'user-new.php' ) …

svenl77:
…to avoid blank admin footer text all around wp.


champsupertramp:
Thank you @svenl77

Issue #173, opened 2016-09-18
No email sent on user password change


vovodroid:
WP 4.6.1

The reason for this: UM do not set $userdata[‘user_pass’] passed to wp_update_user()

https://github.com/ultimatemember/ultimatemember/blob/master/core/um-actions-account.php#L34

$changes is recreated here, and then repopulated again, but all fields with word ‘password’ are not copied from $_POST, thus call to wp_update_user do not causes call to um_send_password_change_email action.

So I added

```
if (isset($_POST[‘user_password’]))
   $changes[‘user_pass’] = $_POST[‘user_password’];
```
after foreach loop (line 40) and sending started to work.

Regards

PR #178, opened 2016-09-19
Fix No email sent on user password change #173

vovodroid:
Fix for https://github.com/ultimatemember/ultimatemember/issues/173


ultimatemember:
Will have our team test this out and see if they can replicate issue. Thanks

PR #179, opened 2016-09-19
Fix error logon message.

vovodroid:
Now there are separate messages for wrong user name and password.
Actually it's not secured, let's return the same message to prevent user name probing.


ultimatemember:
Will discuss this as a team. Thanks.

PR #180, opened 2016-09-19
Add {user_ip} short code

vovodroid:
As it title says )
Allows include user ip (from connection and passed by proxy) to emails.


ultimatemember:
Thanks for the suggestion. We've not had any other requests for this feature so it is not something we are looking to add to the plugin at this time. Thanks :)

PR #183, opened 2016-09-19
Add disable all emails option

vovodroid:
Option to disable call to wp_mail. Useful for site tests.


ultimatemember:
We've not had any other requests for such a feature so will add to user suggestion list for the time being. Thanks :)

PR #185, opened 2016-09-19
Add Show profile on Account page option

vovodroid:
Now on account page photo, name and link View profile are shown.
This is option to disable it, as it can be not suitable for all sites.


ultimatemember:
No plans to hide this as profiles are a required component of UM. Thanks.

PR #189, opened 2016-09-23
Cleanup wp_options table upon user deletion in Wordpress

vovodroid:
Cleanup options like um_cache_userdata_99 in wp_options table.


champsupertramp:
hi @vovodroid, Thanks for all the pull requests. Our team will review these codes. Thank you again.

PR #208, opened 2016-12-01
Correct mobile class on account delete tab heading

tosinafolabi:
.uimob300-hide (instead of .uimob340-hide) applied to element meant the "display:none" styling did not get applied on mobile screens.


champsupertramp:
Thank you

PR #280, opened 2017-06-10
Fix error missing wp-load.php

abdusfauzi:
Fixing bug which require missing wp-load.php
See ultimatemember#7

### DIFF
--- a/core/lib/upload/um-file-upload.php
+++ b/core/lib/upload/um-file-upload.php
@@ -1,12 +1,24 @@
 <?php
 
-$i = 0;
 $dirname = dirname( __FILE__ );
 do {
 	$dirname = dirname( $dirname );
+	$wp_config = "{$dirname}/wp-config.php";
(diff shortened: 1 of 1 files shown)


champsupertramp:
Can you please provide steps to recreate the issue?

abdusfauzi:
WordPress allow us to install the `wp-config.php` in a different folder and custom `wp-content`, such as `core` for WP, `app` for `wp-content` and refer them back in `index.php`. Example:

```
index.php
wp-config.php
/app/uploads/
/app/plugins/
/app/...
/core/wp-admin/
/core/wp-includes/
/core/...
```

The configuration above will turn the root of plugin as `/app/plugins/ultimate-member/` and not `/wp-content/plugins/ultimate-member/` like always. Thus, when the script navigate up the folder structure, they will never find `wp-load.php` which usually at the root, because now it is located in `/core/wp-load.php`.

PR #303, opened 2017-07-28
Fix save and query of privacy settings

malled:
Field hide_in_members was not saved at all. When saving, it was saved translated, i.e. changing the plugin language breaks the feature.


nikitasinelnikov:
Fixed and pushed, Thanks for this fix!

PR #362, opened 2017-12-16
Takes into account the custom order of the options

LuigiPulcini:
The current code always disregards the order the user wants for the roles when customizing the options of role_radio and role_select, because the $options array gets built following the default order of the roles.

With this approach, the custom order is taken into account while building the option list.


nikitasinelnikov:
Thanks! We appreciate it. This fix will be added soon.

PR #378, opened 2018-01-16
Fix errors in um-conditional.js when jQuery.browser is not defined

leolabs:
This PR fixes errors that occur when jQuery.borwser.msie is checked but jQuery.browser is undefined.


leolabs:
Fixes #378

leolabs:
Is now fixed in https://github.com/ultimatemember/ultimatemember/commit/58e422564742707ae810158b00d4d4d7e1abf6ba, so no longer necessary.

PR #432, opened 2018-04-29
Fix conditionals when using multiple forms on one page

leolabs:
When using multiple forms on one page, the conditional script affected fields in one form based on conditionals from other forms on the same page. This commit fixes this faulty behavior.


chroust:
When do you plan to apply this fix to the main branch?

leolabs:
Yeah, it'd be great to have that in the master branch so I can update Ultimate Member again without breaking my site 😄

Issue #484, opened 2018-07-27
Issue in post category role filtering when post has more than one category


jdursel:
### Ultimate Member Version
Tell us what UM core version you use. 
- [x] 1.3.88 ( latest )
- [ ] 2.0 ( beta )

### Subject of the issue
When a post has more than one category of type "logged in" (_um_accessible = 2), the first category of that kind is always used. When a user hasn't the roles matching that category, the post is filtered out. And user cannot access it.

### Expected behavior
When the post has more than one (_um_accessible=2) category, all thoses categories must be tested until one is matching the current user roles.
On the first post category role matching, that restriction rule must be returned. If none of them matches, the last restriction rule can be returned (the user hasn't access).

### Actual behavior
When a post has multiple category for a logged in user, it uses the first encountered rule and if the user don't have the corresponding role, he is not allowed to see the content of the post (although he has at least one matching role of the category)

### Steps to reproduce the behavior
- install ultimate-member 
- create 3 roles: good, bad, ugly
- create a new user with the "bad" role (optionnaly create the "good" user with the "good" role and the "ugly" with the "ugly" role)
- create a new post "gold quest" with the following roles: "good" and "bad" 
- login with the "bad" user
- when you go check the posts, you cannot see the "gold quest"post even if the quest is for the good and the bad (abnormal)
- (optional) if checked with the "good", he has access (normal)
- (optional) if checked with the "ugly", he hasn't access (normal)

## PHP version
- [x] 7.1 above
- [ ] 5.4 
- [ ] 5.3 below

### Do you use UM extensions? 
- [ ] Private Content 
- [ ] Instagram
- [ ] User Tags 
- [ ] Social Activity
- [ ] WooCommerce
- [ ] Private Messages
- [ ] Followers
- [ ] Real-time Notifications
- [ ] Social Login
- [ ] bbPress
- [ ] MailChimp
- [ ] User Reviews
- [ ] Verified Users
- [ ] myCRED
- [ ] Notices
- [ ] Profile Completeness 
- [ ] Friends
- [ ] Terms & Conditions
- [ ] Google reCAPTCHA
- [ ] Online Users

jdursel:
Here is the solution to this issue, but I don't know how to self-assign this issue...
Here's the solution:
In ultimate-member/includes/core/class-access.php, in function get_posts_privacy_settings( $post ), replace the last "foreach" by this code (line 574 to line 588):
`
			//get restriction options for first *matching* term with privacy settigns
[...]

PR #486, opened 2018-07-30
Fix role filtering for posts with multiple categories

jdursel:
Fix for #484 
The filter will retrieve the first matching filtering that authorizes the current user to see the post.
If no terms are matching, then return the last restriction rule.


denisbaranov:
This bug was fixed. Current version of the Ultimate Member works fine.

PR #530, opened 2018-10-03
Fix JS um_select_options_cache overlap

pierre-r:
When multiple fields depend on the same parent, cache is overlaping from one to another (the slowest request is assigned to all child fields).
This code fix it, to save cache dependings on the child ID and parent value, and not only parent value


pierre-r:
I've added a filter to select fields options.
I noticed something strange in the code. Sometimes when you set the options, you use something like:
`foreach($options AS $k => $v) 
echo '<option value="'.$k.'">'.$v.'</option>';`
And sometimes, you define the options as:
`foreach($options AS $k => $v) 
echo '<option value="'.$v.'">'.$v.'</option>';`

It's prety annoying, especially if you combine this with the ACF plugin (Advanced Custom Fields).

By adding this filter, I can write a function that force the option value to be always the same.

Here is an exemple of what I do in my theme:
`function my_fields_options_format_pairs($options)
{
    foreach ($options AS $k => $v) {
        $options[$v] = $v;
        unset($options[$k]);
    }
    return $options;
}

add_filter('um_select_options_format_pair', 'my_fields_options_format_pairs', 10, 2);`

PR #671, opened 2020-09-13
Fix: This is not possible for security reasons

sarojlakra:
This change solves the "_This is not possible for security reasons_" issue and properly checks for the invalid Role entry during registration form submit.
**in_array()** check has a problem. Its comparing existing **$custom_field_roles** array value like um_customrole to CustomRole.
This update fixes **$custom_field_roles** values before check.

Thank you.


sarojlakra:
Edit: These changes should be in the function custom_field_roles( )

PR #709, opened 2020-12-15
Indexing of Profiles

denisbaranov:
New user role setting "Avoid indexing profile by search engines" is added. This setting sets the default value for the account setting "Avoid indexing my profile by search engines".

UM User Roles, Edit Role + Avoid indexing profile by search engines

**Proposition 1**
I think it would be better to remove the account setting "Avoid indexing my profile by search engines". Users can avoid indexing by making their profiles private. Users may be confused by seeing a lot of similar settings.

UM Account page, Profile Privacy + Avoid indexing my profile by search engines

**Proposition 2**
I think it would be better to change the logic to opposite - disable indexing by default and change the user role setting "Avoid indexing profile by search engines" to the "Allow indexing profile by search engines".


nikitasinelnikov:
@denisbaranov please add this ability also:

"If we're able to set a default value on the backend"

https://www.screencast.com/t/Zp27skYsv - global setting to using "noindex". The role setting has priority.

denisbaranov:
The general setting "Avoid indexing profile by search engines" is added

PR #724, opened 2021-01-07
Show button to generate default pages if lost.

denisbaranov:
Changes:
1) Show a notice and a button to generate default pages if one of the pages is lost.
2) New notice _'Ultimate Member Setup Error: Account page and User page should be separate pages.'_
3) Display the empty option _'Select...'_ for pages that were not selected.
4) Clear a list of hidden admin notices on the core plugin re-activation.


nikitasinelnikov:
Already fixed in the 033b6fd167d2ad4debc1e1916c02de4a4270f414 commit

PR #734, opened 2021-01-22
The empty option show in field

denisbaranov:
Unselects empty option in the multi-select field if some other option is selected

The example of the issue:
Empty option


nikitasinelnikov:
Manually merged to avoid the conflicts

PR #743, opened 2021-02-07
Set the locale to the locale of the email recepient

markokhman:
Currently when using the Ultimate Member with together with WPML plugin there is an issue with emails that are sent to the user of the website. Let me describe a case:

A new user registers from a page in German language and the locale of this user is set to German. When admin approves his account via Ultimate Member - if the admin's locale is English, then the notification about the account approval is sent in English language, which is not acceptable, as the user is obviously German speaking. 

I have implemented a following solution, which seems quite ugly, but it works for me. Please let me know if we can somehow add this kind of code into the Ultimate Member to avoid such an issue.

### DIFF
--- a/includes/core/class-mail.php
+++ b/includes/core/class-mail.php
@@ -405,6 +405,10 @@ if ( ! class_exists( 'um\core\Mail' ) ) {
 				return;
 			}
 
+			$user = get_user_by( 'email', $email );
+			$user_locale = get_user_meta( $user->ID, 'user_locale', true );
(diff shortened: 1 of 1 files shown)


nikitasinelnikov:
Hi @markokhman 

Thanks for your pull request and suggestions! 

Since the integration with WPML can have different implementation paths and logic on websites, we added hooks before and after sending email notifications. This will help you implement the language switching logic you need. Please remember to switch the language back after submitting.

These hooks will be added in 2.1.17 version.

Best Regards!

PR #745, opened 2021-02-07
Issue with checkboxes not setting as selected on custom fields checkboxes

markusila:
I have created custom fields - checkboxes.
But when i am at editing profile page - the checkboxes are not shown as set after they are set. Please have a look at this video to reproduce:
https://www.loom.com/share/d2f7788fff16415f9c88720e7944746d


nikitasinelnikov:
Hi @markokhman 

Could you clarify what Ultimate Member and Ultimate Member - Profile Tabs versions do you use?
I cannot see the issue with checkboxes in UM 2.1.17-alpha and UM: Profile Tabs 1.0.2.

Let me know,
Thanks!

Issue #757, opened 2021-02-23
Missing Password Changed confirmation email notification after reseting password via password reset page.


alexpull:
**Expected behavior**

The Password Changed confirmation email notification should be sent to users who reset their password via "Forgot your password" -> Password reset page. Currently, this notification only works if users reset their password via the Account page -> Change password section.

It gives feedback to the user that their action ( to reset their password ) has actually happened. It alerts people that their password has changed in case they didn't request it.

**Isolating the problem (mark completed items with an [x]):**
- [x ] I have deactivated other plugins and confirmed this bug occurs when only Ultimate Member plugin is active.
- [x] This bug happens with a default WordPress theme active, or [UM Theme](https://ultimatemember.com/theme/).
- [x] I can reproduce this bug consistently using the steps above.

PR #761, opened 2021-02-24
Missing Password Changed confirmation email notification after reseting password via password reset page.

yuriinalivaiko:
Sends the "Password Changed Email" when the password is changed via password reset page.


nikitasinelnikov:
#757 is fixed

PR #778, opened 2021-03-12
Changes for user locations filters

ashubawork:
- added hooks for change a wp_um_metadata table
- added hooks for user_location fields (hooks for comparison in text fields)

Pull request for user locations here - https://github.com/ultimatemember/user-locations/pull/5

### DIFF
--- a/includes/core/class-member-directory-meta.php
+++ b/includes/core/class-member-directory-meta.php
@@ -74,7 +74,7 @@ if ( ! class_exists( 'um\core\Member_Directory_Meta' ) ) {
 				update_option( 'um_usermeta_fields', array_values( $metakeys ) );
 			}
 
-			apply_filters( 'um_metadata_on_delete_custom_field', $metakeys, $metakey, $args );
+			do_action( 'um_metadata_on_delete_custom_field', $metakeys, $metakey, $args );
 		}
 
 
@@ -92,7 +92,7 @@ if ( ! class_exists( 'um\core\Member_Directory_Meta' ) ) {
 				update_option( 'um_usermeta_fields', array_values( $metakeys ) );
 			}
 
-			apply_filters( 'um_metadata_on_new_field_added', $metakeys, $metakey, $args );
+			do_action( 'um_metadata_on_new_field_added', $metakeys, $metakey, $args );
 		}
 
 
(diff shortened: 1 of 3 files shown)


includes/core/class-member-directory-meta.php
```diff
@@ -56,51 +56,25 @@ function __construct() {
...
-							'%s'
-						)
-					);
-
-					update_option( 'um_usermeta_fields', array_values( $metakeys ) );
-				}
+				update_option( 'um_usermeta_fields', array_values( $metakeys ) );
 			}
+
+			apply_filters( 'um_metadata_on_delete_custom_field', $metakeys, $metakey, $args );
```
nikitasinelnikov, line 77: Filter that doesn't filter anything =)

includes/core/class-member-directory-meta.php
```diff
@@ -113,32 +87,12 @@ function on_delete_custom_field( $metakey, $args ) {
...
-				if ( ! in_array( $metakey, $metakeys ) ) {
-					$metakeys[] = $metakey;
-					update_option( 'um_usermeta_fields', array_values( $metakeys ) );
-				}
+			if ( ! in_array( $metakey, $metakeys ) ) {
+				$metakeys[] = $metakey;
+				update_option( 'um_usermeta_fields', array_values( $metakeys ) );
 			}
+
+			apply_filters( 'um_metadata_on_new_field_added', $metakeys, $metakey, $args );
```
nikitasinelnikov, line 95: And the same, maybe use `do_action` instead?

nikitasinelnikov (changes requested):
See my comments in the lines below

includes/core/class-member-directory.php
```diff
@@ -1596,13 +1596,12 @@ function filters( $directory_data ) {
...
 
 									$value = stripslashes( $value );
-
 									$field_query = array(
 										'relation' => 'OR',
 										array(
 											'key'       => $field,
 											'value'     => trim( $value ),
-											'compare'   => 'LIKE',
+											'compare'   => apply_filters( 'um_members_directory_filter_text', '=', $field )
```
nikitasinelnikov, line 1604: Check the reason for using '=' by default instead of 'LIKE'
maybe we need '=' only for some selected fields?

Old logic - filter by the text field makes the search by Like

nikitasinelnikov (changes requested):
Please check and let me know

PR #781, opened 2021-03-15
Field privacy option "Only visible to profile owner and specific roles"

yuriinalivaiko:
Fix: The field is visible for the profile owner if the field privacy option is "Only visible to profile owner and specific roles".

### DIFF
--- a/includes/um-short-functions.php
+++ b/includes/um-short-functions.php
@@ -1514,6 +1514,7 @@ function um_can_view_field( $data ) {
 
 	if ( isset( $data['public'] ) && UM()->fields()->set_mode != 'register' ) {
 
+		$current_user_roles = [];
 		if ( is_user_logged_in() ) {
 			$previous_user = um_user( 'ID' );
 			um_fetch_user( get_current_user_id() );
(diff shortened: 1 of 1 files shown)


includes/um-short-functions.php
```diff
@@ -1551,19 +1551,13 @@ function um_can_view_field( $data ) {
...
-						}
-					} else {
-						if ( ! um_is_user_himself() && ( empty( $current_user_roles ) || ( ! empty( $data['roles'] ) && count( array_intersect( $current_user_roles, $data['roles'] ) ) <= 0 ) ) ) {
-							$can_view = false;
-						}
-					}
+				} elseif ( um_is_core_page( 'user' ) && um_is_user_himself() ) {
+					$can_view = true;
+				} elseif ( empty( $current_user_roles ) || ( ! empty( $data['roles'] ) && count( array_intersect( $current_user_roles, $data['roles'] ) ) <= 0 ) ) {
+					$can_view = false;
```
nikitasinelnikov, line 1560: Please check if the user has  necessary roles and set $can_view = true;

or set $can_view = false; in line 1555 and rewrite some conditions

nikitasinelnikov (changes requested):
Please check my comment

includes/um-short-functions.php
yuriinalivaiko: I don't understand what is wrong:
- It is impossible to check if the user has necessary roles if the user is logged out.
- There is no sense to check if the user has necessary roles if the user is a profile owner. Owners have to see the field no matter what their role is.
- `$can_view = true;` is the default value,

We have to set $can_view = false; if the user is logged out or if logged in user is not a profile owner and has not necessary roles.

yuriinalivaiko:
The code is simplified. The logic has not been changed.
I removed all `$can_view = true;` in if-else operators.

includes/um-short-functions.php
champsupertramp: Hi @yuriinalivaiko  @nikitasinelnikov  Could you also check this privacy issue with the fields? 
https://wordpress.org/support/topic/front-end-40/

PR #796, opened 2021-04-01
Field is throwing hidden error and prevents profile form from saving

yuriinalivaiko:
Fixed: skip invisible required fields on form validation


nikitasinelnikov:
Resolved by #802

PR #800, opened 2021-04-02
Restricted content message is showing twice and removing header

yuriinalivaiko:
Fixed: Restricted Content message does not replace content in the site header and footer.


sam-moreton:
@nikitasinelnikov This fix is not working in UM 2.2.5

PR #835, opened 2021-05-11
Builder avoid autosave

nikitasinelnikov:
remove AJAX handlers, save via hidden field, and sanitized custom field


denisbaranov:
Closed by accident

PR #851, opened 2021-06-11
Content restriction may replace header in Avada theme

yuriinalivaiko:
Fix: Don't replace content in the Avada header.


sam-moreton:
@yuriinalivaiko It looks like this hasn't been merged into the latest release.

PR #992, opened 2022-04-07
Resolve overwrite data value conflict

sourabhgarg01:
Due to the existing code, user is not able to pass any custom value as it is getting overwrite by the default one. So, this update will resolve the conflict.


sourabhgarg01:
Issue #993

nikitasinelnikov:
Closed
https://github.com/ultimatemember/ultimatemember/commit/31e6a0d2dad6e836d42d1ede9061b685c75b24ce

PR #1004, opened 2022-05-13
Custom metadata for profile slug

fahelmoreira:
Add a option to use a custom metadata in the profile slug

### DIFF
--- a/includes/admin/core/class-admin-settings.php
+++ b/includes/admin/core/class-admin-settings.php
@@ -973,8 +973,16 @@ if ( ! class_exists( 'um\admin\core\Admin_Settings' ) ) {
 											'name_dash'  => __( 'First and Last Name with \'-\'', 'ultimate-member' ),
 											'name_plus'  => __( 'First and Last Name with \'+\'', 'ultimate-member' ),
 											'user_id'    => __( 'User ID', 'ultimate-member' ),
+											'field'      => __( 'Custom field(s)', 'ultimate-member' ),
 										),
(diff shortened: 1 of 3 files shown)


ravneetsinghkis:
https://prnt.sc/ritc8Wfh8wou
I merged your code to changes url slug functionality In ultimate plugin but your code not reflecting

ravneetsinghkis:
Ok Thank you for this but I want to use base64 encoding value in url it will work for this

deepukisworks:
I want to change my profile slug with my custom functionality. is  your code will work if I modify your file if yes can you please let me know how it will work I checked with above option is working fine but my URL is not working
https://prnt.sc/W2Mr8jWUtVfc

fahelmoreira:
sure @deepukisworks, but did you create the custom field on the user profile page? and if so did you add the name as I did? btw the "profile_sug" it's just a placeholder for the name needs to be the same name as the field you created

deepukisworks:
No, I just modified my plugin with your code and get results can you please let me know in brief means do i need to create custom field on profile page and what next step I have to follow

nikitasinelnikov:
The ability to use custom user meta as slug is added here:
https://github.com/ultimatemember/ultimatemember/pull/1299

@ravneetsinghkis
There is also an added ability to set the user slug as hash string. Or if you need own encrypt the user slug please use this filter together with the "Custom usermeta" permalink base. https://github.com/ultimatemember/ultimatemember/pull/1299/files#diff-5f2f61ac3e24754e0d12d9b931d25a7daf0a0076acc16299ab3990d719fede20R803

PR #1014, opened 2022-06-05
Fix for Choices Callback option

yuriinalivaiko:
The dropdown field option **"Choices Callback"** may contain backslashes if the needed callback function is defined inside the namespace.

Post meta values are passed through the `stripslashes()` function upon being stored, so we have to use the function `wp_slash()` to compensate for the call to `stripslashes()`.

**See** https://developer.wordpress.org/reference/functions/update_post_meta/

**Example**
example with namespace


nikitasinelnikov:
Closed via fe88b409392be13a8a2bd90e7c0b831ce34c33a7

PR #1018, opened 2022-06-11
Strip tags in plain text email.

yuriinalivaiko:
We must not use HTML in plain text emails.
Current code strips tags from the email template but doesn't strip tags in placeholders.
In this fix I added strip tags operation after email placeholders are replaced with values.

**Example -  Current plain text email "New User Notification"**
```
UM Testing
test email has just created an account on UM Testing.
To view their profile click here: http://um/profile/test.email.03/
Here is the submitted registration form:
<p><label>Date Submitted: </label><span>11 Jun 2022 20:34</span></p><p><label>Form: </label><span>Default Registration - Form ID#: 5</span></p><div class="um-row _um_row_1 " style="margin: 0 0 30px 0;"><p><label>Username: </label><span>test.email.03</span></p><p><label>First Name: </label><span>test</span></p><p><label>Last Name: </label><span>email</span></p><p><label>E-mail Address: </label><span>umdevelopera+test.email.03@gmail.com</span></p><p><label>Mailchimp: </label><span>No</span></p>
Thank you!
The UM Testing Team
```

**Example -  Fixed plain text email "New User Notification"**
```
UM Testing
test email has just created an account on UM Testing.
To view their profile click here: http://um/profile/test.email.04/
Here is the submitted registration form:
Date Submitted: 10 Jun 2022 15:59
Form: Default Registration - Form ID#: 5
Username: test.email.04
First Name: test
Last Name: email
E-mail Address: umdevelopera+test.email.04@gmail.com
Mailchimp: Yes
Thank you!
The UM Testing Team
```


includes/core/class-mail.php
```diff
@@ -384,9 +384,21 @@ function prepare_template( $slug, $args = array() ) {
 			add_filter( 'um_template_tags_replaces_hook', array( &$this, 'add_replace_placeholder' ), 10, 1 );
 
 			// Convert tags in email template
-			return um_convert_tags( $message, $args );
-		}
+			$message = um_convert_tags( $message, $args );
```
nikitasinelnikov, line 387: Add $context="html|plain" variable for the `um_convert_tags` for making the result more expectable and run this function with "plain" context here.

includes/core/class-mail.php
```diff
@@ -384,9 +384,21 @@ function prepare_template( $slug, $args = array() ) {
...
 
+			// Strip tags in plain text email.
+			// Important! Don't use HTML in plain text emails.
+			if ( ! UM()->options()->get( 'email_html' ) ) {
+				$message_nl = str_replace(
```
nikitasinelnikov, line 392: Because there can be not only `<p>` or `<div>` tags where we have to add \n for plain-text formatting

nikitasinelnikov (changes requested):

nikitasinelnikov:
Implemented plain-text folder with plain-text templates:
2d13b2059164c150e592888dd780dfa5fe9e268f

PR #1020, opened 2022-06-11
Empty values are saved to the DB

yuriinalivaiko:
Delete row from the database table **usermeta** if the field value is empty to reduce the number of rows in the table and avoid crashes with a large number of users.

### DIFF
--- a/includes/core/class-user.php
+++ b/includes/core/class-user.php
@@ -1653,9 +1653,11 @@ if ( ! class_exists( 'um\core\User' ) ) {
 		 * @param $key
 		 */
 		function update_usermeta_info( $key ) {
-			// delete the key first just in case
+			// delete the key first just in case.
 			delete_user_meta( $this->id, $key );
-			update_user_meta( $this->id, $key, $this->profile[ $key ] );
+			if ( isset( $this->profile[ $key ] ) && '' !== $this->profile[ $key ] ) {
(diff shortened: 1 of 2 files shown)


MissVeronica:
This solution will only solve your issue when having low perscentage utilization of the usermeta fields and will not address sites with many fields being used to 100%. 

The backend performance must be improved by using caching for the account_status summaries and one redundant select statement has now been removed in UM 2.4.1 hopefully solving one of the backend issues. Much more can be done to reduce response times in the WP backend for sites with large number of users.

nikitasinelnikov:
Hi @yuriinalivaiko @MissVeronica 

I'm not sure that this solution is used and possible to merge for the future version.
I'll try to explain why.

Since the UM plugin has the ability to create fields using metadata keys that can be associated with other plugins, they cannot be removed from the database. This may lead to conflicts with other solutions. In addition, each deletion and addition of metadata (table rows) will increase the auto-increment. A very special case if this increment reaches the limit.

Let's take a standard set of metadata for a user. For some reason, WordPress itself creates an empty `description` entry and doesn't delete it if even it's empty.

So I'll close this PR but there is another one with the fix for the performance issue and wp-admin side https://github.com/ultimatemember/ultimatemember/pull/1026
I have added some changes to it by adding and deducting a number instead of recalculation.

Let me know if you have other question,
Best Regards!

Issue #909, opened 2021-08-30
Ultimate Member Extremely Slow


MissVeronica:
**Is your feature request related to a problem? Please describe.**
Ultimate Member is using "SQL_CALC_FOUND_ROWS" which is a deprecated way of counting rows. MYSQL is about to remove support for “SQL_CALC_FOUND_ROWS” altogether from MYSQL. This MYSQL query is wellknown for being extremely slow and makes UM sites with large number of users like 200.000 and more very slow.

**Describe the solution you'd like**
The SQL Query is used to recalculate the cached number of users after a user status change because the status change is deleting the current cahed value. Adding or subtracting the changed user to the current cached value should make the costly query not necessary any more.

**Describe alternatives you've considered**
Turning off execution of the SQL query with a code snippet.

`add_action( 'um_when_status_is_set', 'my_disable_um_cached_users_queue', 10, 1 );

    function my_disable_um_cached_users_queue( $userid ) {
    
        remove_action( 'um_after_user_status_is_changed_hook', array( UM()->classes['user'], 'remove_cached_queue' ));
    }`

**Additional context**
https://wordpress.org/support/topic/ultimate-member-extremely-slow/

vijayant123:
My WordPress installation has grown to 500,000+ users and has grown to be slower and slower because of this query in Ultimate Member. Ultimate Member is a great plugin that fits in perfectly for our requirements. But even with a dedicated MySQL 8 core 16 GB RAM VM on AWS these queries involving SQL_CALC_FOUND_ROWS take anywhere from 40 to 90 seconds depending on active server load. This leads to very poor user experience. In the admin dashboard, following pages refuse to load even under a mild user load:
1. All users
2. All pages
3. All posts

I have been using a page timeout of 120 seconds and a very costly MYSQL server to overcome this problem in Ultimate Member plugin’s code. As I mentioned at the start this needs to be fixed ASAP as the SQL_CALC_FOUND_ROWS has already been deprecated and might stop working in any future version of MYSQL. Please look into this with high priority as it will break websites across the world if not fixed timely. This long query problem is easily detectable when the users and usermeta tables grow to significant sizes.

BadOmn83:
I wanted to follow up. After doing some tests and looking at the Dev Tools in Chrome. It seems like calls to admin-ajax.php are what is slowing my site down when loading directory pages. Perhaps too many calls are being made?

MissVeronica:
New performance issue in the UM support forum
[...]

PR #1021, opened 2022-06-13
UM Account status caching 1

MissVeronica:
Required UM updates for an account status caching plugin under development to improve performance for sites with high number of users and/or many meta fields.

This filter hook will return current cached value for this account status:
`$status_count = apply_filters( 'account_status_caching_filter', false, $status );
`
This action hook will update current cached value with counter for this account status:
`do_action( 'account_status_caching_action', $status_count, $status );
`

**Plugin Functions**

The Plugin is using two new Hooks in UM for bypassing the account_status summaries being updated from the wp_usermeta table each time WP All Users and UM Dashboard page displays.

The Pending Account statuses for Approval and e-mail Verification summary for the WP Users button red dot number is calculated from actual cache values.

Cache totals are updated when an user is getting a status change like Approved by an Admin or User Validation of an email address. New registrations are added to cache totals. Existing UM Hooks are being used and two hooks executing the remove_cached_queue function are disabled.

The UM function of approving unassigned user statuses is being set to execute after a time interval of 3600 seconds instead of each new WP Users All page display. If found unassigned users the Approved account_status cache is updated.

An Account Status Cache metabox is added to the UM Dashboard page where the Admin can Refresh the cache totals if required.


MissVeronica:
See also the issue about "Ultimate Member Extremely Slow": https://github.com/ultimatemember/ultimatemember/issues/909

nikitasinelnikov:
Hi @MissVeronica 

There is a performance fix with using transients. I guess it's ok for the performance:
https://github.com/ultimatemember/ultimatemember/pull/1026

Let me know if you have questions,
Best Regards!

MissVeronica:
@nikitasinelnikov 

I have a beta plugin using options table but transients is probably OK also.

You never take advantage of status change hooks for increasing or decreasing the current transient values. 
As I can see you are flushing the transient when an account status is updated instead of getting the transient for update and save new value. Transient flush should only be performed by an admin "Clear account cache" function.

These useful hooks are: `'um_when_status_is_set'`,` 'um_after_user_status_is_changed_hook'`, `'um_delete_user'`.
They cover new user, delete user, email verifications and approvals.

nikitasinelnikov:
Thanks for your prompt reply @MissVeronica 

Please see these hooks and callbacks for them. https://github.com/ultimatemember/ultimatemember/pull/1026/files#diff-5f2f61ac3e24754e0d12d9b931d25a7daf0a0076acc16299ab3990d719fede20R113-R115

We bind at the lowest level during the meta_value change (`add_user_meta`, `update_user_meta` or `delete_user_meta`).
So don't need to hooked every function when user account status changed.
On 'um_delete_user' and 'user_register' hooks there are callbacks for flushing 'um_count_users_unassigned' transient.

See https://github.com/ultimatemember/ultimatemember/pull/1026/files#diff-5f2f61ac3e24754e0d12d9b931d25a7daf0a0076acc16299ab3990d719fede20R393
And https://github.com/ultimatemember/ultimatemember/pull/1026/files#diff-5f2f61ac3e24754e0d12d9b931d25a7daf0a0076acc16299ab3990d719fede20R650

I agree that we can drop transients on the "Clear account cache" button click via wp-admin dashboard: see the latest commit
https://github.com/ultimatemember/ultimatemember/pull/1026/files#diff-ea9dcc68dc045410769870752b2d13527050c8197c29df5cc2a6097c1edf4cc5R1676-R1689

Let me know you thoughts,
Thanks!

MissVeronica:
@nikitasinelnikov

OK but the result in all flush functions is a `delete_transient` which is my objection.

If you have a before and after value for an account_status counter you should decrease one transient and increase another transient by one. 

Now this flush function implies a new database search for the number of each of the updated account_status  values: Poor performance

nikitasinelnikov:
Hi @MissVeronica 

If you have the opportunity to verify that your proposals in this PR and  this https://github.com/ultimatemember/ultimatemember/pull/1022 had been integrated via this PR https://github.com/ultimatemember/ultimatemember/pull/1026 it will be good

I hope that latest changes help to reduce the load on sites with a large number of users.

Let me know,
Best Regards!

MissVeronica:
@nikitasinelnikov

When I see that `delete_transient( "um_count_users_{$value}" );` is being replaced by code doing `$count--;` and `$count++;` followed by calls like `set_transient( "um_count_users_{$old}", $count );` it's verified for me that you have understood my objections.

One additional request might be an "account status clear cache" function in the dashboard deleting all the transients, for the admins to force an immediate controlled recalculation of all the cached status values.

nikitasinelnikov:
Hi @MissVeronica 

Thanks for verifying! Yes, sure, by default we moved the cache reset to the already created metabox on the Ultimate Member > Dashboard screen. See here https://www.screencast.com/t/MPFqNygEtKF

Maybe it's not fully clear, but I can add some texts for the details.

Best Regards!

MissVeronica:
@nikitasinelnikov

An additional metabox only for the new account cache reset would best solution.

nikitasinelnikov:
@MissVeronica 
A separate button has been added for that.
Thanks for your suggestions!

Issue #988, opened 2022-03-21
Make Mobile/Phone Number UM Fields type="tel" and Clickable


MissVeronica:
**Is your feature request related to a problem? Please describe.**
With the current type="text" no integrations with Mobile/Phone support Plugins are possible like geolocation/country code.
Entering a Mobile/Phone Number on a Mobile is not using the Mobile Number Keyborad with type="text".

**Describe the solution you'd like**
UM Fields for Moble/Phone numbers to generate type="tel" fields both in Registration and Profile editing.
Option of using HTML5 pattern attribute with a regular expression validation of the number format.
Display the href="tel:" format of Mobile/Phone fields to generate clickable links for calling the number on mobiles.

**Describe alternatives you've considered**
https://github.com/MissVeronica/UM-Mobile-Number-Login/blob/main/add-on-1.php
https://github.com/MissVeronica/UM-Mobile-Number-Login/blob/main/add-on-2.php

**Additional context**
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/tel

Mobile Number Login https://github.com/ultimatemember/ultimatemember/issues/974

PR #1033, opened 2022-06-29
Phone fields (issue #988)

ashubawork:
- change type for phones fields (issue #988)

### DIFF
--- a/includes/core/class-builtin.php
+++ b/includes/core/class-builtin.php
@@ -245,6 +245,22 @@ if ( ! class_exists( 'um\core\Builtin' ) ) {
 					)
 				),
 
+				'tel' => array(
+					'name' => 'Tel Box',
+					'col1' => array('_title','_metakey','_help','_default','_min_chars','_visibility'),
+					'col2' => array('_label','_placeholder','_public','_roles','_validate','_custom_validate','_max_chars'),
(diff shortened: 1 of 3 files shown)


MissVeronica:
There is an issue with the href statement with `tel` in `function um_profile_field_filter_hook__phone`:       

Remove the equal sign in `href="tel:=' . `

Code should be:
`$value = '<a href="tel:' . esc_attr( $value ) . '" rel="nofollow" title="' . esc_attr( $data['title'] ) . '">' . esc_html( $value ) . '</a>';`

See this example: https://www.w3schools.com/tags/tryit.asp?filename=tryhtml_link_phoneto

MissVeronica:
Second issue in this `function um_profile_field_filter_hook__phone`

You can't remove the + sign if present in the mobile number:  `$value = str_replace('+', '', $value);`
It's user or a plugin responsibility to add the + sign in international numbers with country codes.

The + sign is detected by the mobile phone which will add the exit code for the country where the mobile is located, otherwise the number is considered being local for the country.

Source:   https://www.howtocallabroad.com/qa/plus-sign.html

Plugin example:    https://ps.w.org/international-telephone-input-with-flags-and-dial-codes/assets/screenshot-2.png?rev=2465547

PR #1058, opened 2022-08-05
Update error message - unique_email

0xsuid:
Updated error messages for 3 conditions:

1. when user already exists - registration mode/form 
    - "The email you entered is incorrect" to "The email you entered is used by another user"
3. when user already exists - profile mode/form
    - "The email you entered is incorrect" to "The email you entered is used by another user"
5. Invalid email
    - "The email you entered is incorrect" to "The email you entered is incorrect"

### DIFF
--- a/includes/core/um-actions-form.php
+++ b/includes/core/um-actions-form.php
@@ -811,11 +811,11 @@ function um_submit_form_errors_hook_( $args ) {
 								if ( $args[ $key ] == '' && in_array( $key, array( 'user_email' ) ) ) {
 									UM()->form()->add_error( $key, __( 'You must provide your email', 'ultimate-member' ) );
 								} elseif ( in_array( $mode, array( 'register' ) ) && $email_exists  ) {
-									UM()->form()->add_error( $key, __( 'The email you entered is incorrect', 'ultimate-member' ) );
+									UM()->form()->add_error( $key, __( 'The email you entered is used by another user', 'ultimate-member' ) );
 								} elseif ( in_array( $mode, array( 'profile' ) ) && $email_exists && $email_exists != $args['user_id']  ) {
-									UM()->form()->add_error( $key, __( 'The email you entered is incorrect', 'ultimate-member' ) );
+									UM()->form()->add_error( $key, __( 'The email you entered is used by another user', 'ultimate-member' ) );
 								} elseif ( ! is_email( $args[ $key ] ) ) {
-									UM()->form()->add_error( $key, __( 'The email you entered is incorrect', 'ultimate-member') );
+									UM()->form()->add_error( $key, __( 'Please enter valid email address', 'ultimate-member') );
 								} elseif ( ! UM()->validation()->safe_username( $args[ $key ] ) ) {
 									UM()->form()->add_error( $key,  __( 'Your email contains invalid characters', 'ultimate-member' ) );
 								}


MissVeronica:
Why should UM switch back to the mode where registered user's email can be disclosed by scammers?
This security issue has been resolved during the last year by UM.

nikitasinelnikov:
Hi @0xsuid 

Thanks for your pull request, but it cannot be added back to Ultimate Member based on security reasons. We also don't recommend making these changes locally on your website.

If you are ready to endanger your users, you can take advantage of changing these texts with a string translation plugin.

Best Regards!

0xsuid:
Hello @nikitasinelnikov @MissVeronica,

I have sent this PR because few users faced issues while signing up with existing email and they didn't understood error messages, how genuine users will know that their email is already used in system???

As you have told this matter regarding the security, I have taken overview of the plugin in matter of signup and it lets user signup when email is not used and shows **meaningless** error when it's used. one could easily design bot for proof of concept based on this error messages.

Blocking bots should be job of firewall like "wordfence" or better logic which prevents bots on registration bots, just by showing meaningless errors it is not stopping bots.

In nutshell, you are showing any kind of error when email is used, so it can be used by bots to detect that email is used, I would recommend to implement something better which prevent bots and also let user know what they are doing something wrong.

MissVeronica:
Security reasons are not only bots but also includes personal data and GDPR.

What is personal data? 
Email addresses are personal data and personal data is subject to the protection requirements set out in the GDPR.

https://ec.europa.eu/info/law/law-topic/data-protection/reform/what-personal-data_en

With a registration or password reset response that an email address is registered on a site you have revealed personal data.

0xsuid:
@MissVeronica By showing error, it just indirectly indicates that account exist on the website. Why it's hard to understand?? 

However, after considering your points of GDPR and security reasons, it would be better if this error messages are removed completely, then and only then those points could be justified. Just showing different message doesn't justify anything.

If you want me remove error messages when email address has been already used then let me know, I would update my PR.

MissVeronica:
The same issue with IP addresses representing personal data because it's theoretically possible to identify the person associated with an IP address.

Website fined by German court for leaking visitor's IP address via Google Fonts

https://www.theregister.com/2022/01/31/website_fine_google_fonts_gdpr/

nikitasinelnikov:
Hi @0xsuid @MissVeronica 

Thank you for the active discussion regarding this issue. Hope this helps end users and developers in the future.

Undoubtedly native WordPress forms for login and password reset display detailed errors (email already exists or not). This can be regarded as an advantage for the end user, and as a disadvantage for the security of the system as a whole.

**The laws of the country in which the end user does business are always important**

Our plugin is used by people of different countries and nationalities and the laws of each country are individual. Somewhere data privacy is a high priority, somewhere less. But for our development team, the main value is the security of the end-user of our plugins. Security issues are always a top priority.

Our plugin, like WordPress as a whole, supports a convenient customization structure for both entire functional blocks and individual texts. In the first case, you can use WP hooks, in the second case you can use textdomain translations.

You can have these customizations in the separate plugin or child-theme or translations file on your end website.

Thank you @0xsuid for your patch and @MissVeronica for your links with GDPR privacy policies.
This patch cannot be applied to the Ultimate Member core due to security vulnerabilities and disclosure of users' credentials on the site. But you can safely apply this patch on your site if it is provided for by the laws of the country where you do business.

Best Regards!

0xsuid:
@MissVeronica @nikitasinelnikov I don't think you got my point, I repeat my point - that by showing any kind of error("The email you entered is incorrect") when user input their email(which already exists in system) in registration form, it just **indirectly indicates** that account with email **exist** on the website. 

If you remove that error message for existing email then and only then your point with GDPR could be justified, otherwise in my humble opinion it is not compliant with GDPR because you have shown a message "The email you entered is incorrect" which is not that hard for human to understand that email is already used.

In the end, not interested in further discussion because we are going round in circles.

Issue #1118, opened 2023-01-09
Birthdates are not UTC dates/times


MissVeronica:
**Expected behavior**
UM is converting the birth date as if the user entered the UTC birth date.
Birth dates are always in local time and only a display format modification is required.

**UM Support Forum issue**
https://wordpress.org/support/topic/incorrect-date-time-in-birthdate/

**Solution**
```
remove_filter( 'um_profile_field_filter_hook__date', 'um_profile_field_filter_hook__date', 99, 2 );
add_filter( 'um_profile_field_filter_hook__date', 'um_profile_field_filter_hook__date_update', 99, 2 );

function um_profile_field_filter_hook__date_update( $value, $data ) {
	if ( ! $value ) {
		return '';
	}
	if ( isset( $data['pretty_format'] ) && $data['pretty_format'] == 1 ) {
		$value = UM()->datetime()->get_age( $value );
	} else {
		$format = empty( $data['format_custom'] ) ? $data['format'] : $data['format_custom'];
		if ( in_array( $data['metakey'], array( 'birth_date' ) ) ) {
			$value = date( $format, strtotime( $value ) );
		} else {
			$value = date_i18n( $format, strtotime( $value ) );
		}
	}

	return $value;
}
```

nikitasinelnikov:
This PR should close this issue
https://github.com/ultimatemember/ultimatemember/pull/1327

PR #1128, opened 2023-01-23
Birthday if timezone is not UTC

ashubawork:
fix the birthday if the timezone is set with date_default_timezone_set() and is not UTC (issue https://github.com/ultimatemember/ultimatemember/issues/1118)

### DIFF
--- a/includes/core/um-filters-fields.php
+++ b/includes/core/um-filters-fields.php
@@ -291,14 +291,7 @@ function um_profile_field_filter_hook__date( $value, $data ) {
 		$value = UM()->datetime()->get_age( $value );
 	} else {
 		$format = empty( $data['format_custom'] ) ? $data['format'] : $data['format_custom'];
-		if ( 'UTC' !== date_default_timezone_get() ) {
-			$site_timezone = date_default_timezone_get();
-			date_default_timezone_set( 'UTC' );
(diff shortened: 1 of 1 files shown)


nikitasinelnikov:
Please check how it works with WordPress default and recommend 
https://developer.wordpress.org/reference/functions/wp_date/
https://make.wordpress.org/core/2019/09/23/date-time-improvements-wp-5-3/

includes/core/um-filters-fields.php
```diff
@@ -291,7 +291,15 @@ function um_profile_field_filter_hook__date( $value, $data ) {
...
 		$format = empty( $data['format_custom'] ) ? $data['format'] : $data['format_custom'];
-		$value = date_i18n( $format, strtotime( $value ) );
+		if ( 'UTC' !== date_default_timezone_get() ) {
+			$site_timezone = date_default_timezone_get();
+
+			date_default_timezone_set( 'UTC' );
+			$value = wp_date( $format, strtotime( $value ) );
+			date_default_timezone_set( $site_timezone );
+		} else {
+			$value = wp_date( $format, strtotime( $value ) );
```
nikitasinelnikov, line 301: Leave only this line. It should be enough

nikitasinelnikov:
Closed by PR https://github.com/ultimatemember/ultimatemember/pull/1327

PR #1145, opened 2023-03-05
- removed predefined field Google+

yuriinalivaiko:
[Google+](https://plus.google.com/) is no longer available for consumer (personal) and brand accounts.
The predefined field **Google+** in the **Fields Manager** has to be removed since it is useless now.

### DIFF
--- a/includes/core/class-builtin.php
+++ b/includes/core/class-builtin.php
@@ -949,24 +949,6 @@ if ( ! class_exists( 'um\core\Builtin' ) ) {
 					'match' => 'https://linkedin.com/',
 				),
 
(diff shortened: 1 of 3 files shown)


MissVeronica:
A remove of the Google+ field should also remove all records in the DB table wp_usermeta with the `metakey` equal to `googleplus`

yuriinalivaiko:
> A remove of the Google+ field should also remove all records in the DB table wp_usermeta with the `metakey` equal to `googleplus`

Some members may still have this field in their profiles. We should not delete user data without permission. After removing the **Google+** field these data will be displayed as a custom **URL** field.

PR #1204, opened 2023-06-09
Fixed wp_editor scripts on macOS

nikitasinelnikov:


ashubawork (approved):
With this fix the wp-editor works well on the macOS in the firefox

PR #1285, opened 2023-08-22
Profile form fields disabled in edit mode

yuriinalivaiko:
Fixed the field "Can user edit this field?" (editable) setting:
- make a new field editable by default.
- check that the `editable` setting exists before use it.

### DIFF
--- a/includes/core/class-fields.php
+++ b/includes/core/class-fields.php
@@ -3626,7 +3626,7 @@ if ( ! class_exists( 'um\core\Fields' ) ) {
 								$class  = 'um-icon-android-radio-button-off';
 							}
 
-							if ( empty( $data['editable'] ) ) {
+							if ( array_key_exists( 'editable', $data ) && empty( $data['editable'] ) ) {
 								$col_class .= ' um-field-radio-state-disabled';
 							}
 
@@ -3745,7 +3745,7 @@ if ( ! class_exists( 'um\core\Fields' ) ) {
 							$class  = 'um-icon-android-checkbox-outline-blank';
 						}
 
-						if ( empty( $data['editable'] ) ) {
+						if ( array_key_exists( 'editable', $data ) && empty( $data['editable'] ) ) {
 							$col_class .= ' um-field-radio-state-disabled';
 						}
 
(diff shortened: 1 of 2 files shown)


MissVeronica:
Disabled HTML classes in `class-fields.php` lines 3629 and 3748 will not be updated by this pull request?

yuriinalivaiko:
> Disabled HTML classes in `class-fields.php` lines 3629 and 3748 will not be updated by this pull request?

It was decided to consider the field without the `editable` setting as non-editable. Just fixed the checkbox "Can user edit this field?" in the field settings. See 8cb6c79

MissVeronica:
> It was decided to consider the field without the editable setting as non-editable.

You can't define these fields as non-editable!!!
I have editable fields like this https://imgur.com/a/MkPXslb
which have this format in the `um_can_edit_field` and `um_is_field_disabled` filter's parameter `$data` 
where `disabled` is `false` and `editable` is missing.

```
Array
(
    [classes] =>  um-field-timezone um-field-text um-field-type_text
    [conditional] => 
    [default] => 
    [disabled] => 
    [in_column] => 3
    [in_group] => 
    [in_row] => _um_row_1
    [in_sub_row] => 1
    [input] => text
    [label] => Timezone
    [metakey] => timezone
    [placeholder] => 
    [position] => 82
    [public] => 1
    [required] => 
    [title] => Timezone
    [type] => text
    [validate] => 
    [visibility] => all
)

```

nikitasinelnikov:
Hi @MissVeronica 

How did you get the fields without an "editable" attribute? We need to standardize the data format based on WPCS requirements in UM code.
I need to know the case and standardize the data in DB for this attribute. So I need to know what legacy structure was and when such fields can be created.

Please let me know, I hope it's possible to make it automatically if there are not any logical forks with this data.
Best Regards!

nikitasinelnikov:
@MissVeronica 

Having analyzed the old code, I can unfortunately say that it will be necessary to return the checks like:

`array_key_exists( 'editable', $data ) && empty( $data['editable'] )`

Because of `editable` === null. That doesn't exist in the custom field data equals to `editable` = true.

Anyway after the first field update the custom field data will have the bool value false|true because of necessary sanitizing.
You may ignore my previous comment.

Best Regards!

PR #1289, opened 2023-08-26
Aria accessibility attributes on form errors

TheNthDev:
* Adding aria-errormessage and aria-valid tags to enhance 508 Compliance on UM forms
* Merging from development/2.6.11
* Minor fix in development/2.6.11 that caused error messages to not show up (class-form->processing was changed to an int [class-form.php 713], so === was failing)


TheNthDev:
This is my first attempted PR to a public Github repo, so please let me know if I have done anything incorrectly and I will work to fix it 🙂

PR #1293, opened 2023-08-27
Wrong fields in the login form displayed on the User page

yuriinalivaiko:
Fixed login form fields. 
Get fields from the form arguments if the `custom_fields` key is not empty. Use the default method to get fields otherwise.


nikitasinelnikov:
Fixed by https://github.com/ultimatemember/ultimatemember/pull/1282/commits/a2b52cdd0cec283bb0ddcb7627bcc9a63115c09e

Issue #1301, opened 2023-09-13
Add 'ID' key to blacklist when create the custom field


nikitasinelnikov:
If there is a custom field on the UM form with the metakey=ID then form submission is break.

PR #1302, opened 2023-09-13
Fields key blacklist

ashubawork:
- add the ID to the blacklist of meta key (issue #1301)


nikitasinelnikov:
1. Please make a separate validation type for the meta key field. (Not in "unique" type).
2. Please use the new format of the hookdocs

PR #1306, opened 2023-09-20
Conflict with the Bookly plugin

yuriinalivaiko:
Fixed conflict with the Bookly plugin. Fields in the Form Builder can be edited.


nikitasinelnikov:
Related to Bookly plugin:
https://wordpress.org/support/topic/conflict-with-ultimate-member-plugin-5/

Issue #1325, opened 2023-10-09
YouTube URL validation


nikitasinelnikov:
**Is your feature request related to a problem? Please describe.**
YouTube field type supports YouTube Shorts links. It works with the Short URL but there's no validation if the short URL has valid structure. This short URL is correct: https://www.youtube.com/shorts/RMpHqJdXu80 but when I modified the URL into https://www.youtube.com/shossrts/RMpHqJdXu80, the form submission accepted it showing the embeded non-playable video with the invalid URL.

**Describe the solution you'd like**
[Add YouTube URL validation.](https://imgur.com/euczTnU)

nikitasinelnikov:
Closed but can be enhanced via https://github.com/ultimatemember/ultimatemember/issues/1331

PR #1326, opened 2023-10-09
Youtube URL validation

ashubawork:
- add a YouTube URL validation for video and shorts (issue #1325)


champsupertramp (commented):
Hi @ashubawork This looks good. Maybe in the future, we can use the built-in   WP_oEmbed::get_provider( $url ); class method to validate media URLs:  
Please see: https://github.com/WordPress/wordpress-develop/blob/6.3/src/wp-includes/class-wp-oembed.php#L263-L292

The class already covers the discovery of URL providers with Regex lists:
https://github.com/WordPress/wordpress-develop/blob/6.3/src/wp-includes/class-wp-oembed.php#L54-L57

ashubawork:
Hi @champsupertramp 
Thank you for your feedback. I'll do some research on the links you gave me.

PR #1336, opened 2023-10-18
Cropper destroy

ashubawork:
- fix cropper destroy


nikitasinelnikov:
Cropper init and destroy is here in refactored scripts:
https://github.com/ultimatemember/ultimatemember/tree/development/admin_assets_refactoring

PR #1340, opened 2023-10-27
Fix date and time to filter suspicious accounts + minor fixes to the timestamp in GMT

champsupertramp:
There's an issue with the security scanner. It doesn't filter all suspicious accounts and timestamps are incorrectly saved that caused invalid datetime differences.


champsupertramp:
You may also check this ticket related to this issue:
https://secure.helpscout.net/conversation/2401687410/75165?folderId=670790

PR #1346, opened 2023-10-29
YouTube Video field does not support all YouTube links

yuriinalivaiko:
The YouTube Video field will support YouTube links with additional parameters.


yuriinalivaiko:
This fix is included in pull request #1347.

Issue #1308, opened 2023-09-20
The {submitted_registration} placeholder returns UTC time


MissVeronica:
**Expected behavior**
The `{submitted_registration}` placeholder should return  local time.

**Describe the bug**
The Submitted Registration Unix timestamp is set by `current_time( 'timestamp' );` 
and displayed in human time format with the `wp_date` function which is UTC time.

At the same time GDPR timestamp is set by the `time()` function
and displayed in human format with the `wp_date` function which is formatted local time.

**Expected behavior**
Use `current_time( 'timestamp' )` together with the `date_i18n` function to get local time in human format.

The `time()` function and `wp_date` will give local time and the `date_i18n` function will give an invalid human time.

**Screenshots**
https://imgur.com/a/3USRulq

**Support Forum issue**
https://wordpress.org/support/topic/emails-displaying-wrong-time-zone/

PR #1358, opened 2023-11-08
Submitted registration date

ashubawork:
- Fix the {submitted_registration} placeholder local time (issue #1308)


nikitasinelnikov:
We start to use user_registered date instead of timestamp since 2.8.0

Issue #737, opened 2021-01-27
Content Editor field is not rendering breaking lines correctly .


champsupertramp:
**Is your feature request related to a problem? Please describe.**
The editing the content in the Content Editor field is not rendering breaking lines correctly . 

**Describe the solution you'd like**
For instance, if you press "Enter" key, it should render `<p>` or `<br>` but it's just rendering whitespace. Once you've saved the Content Block and view it in the Profile form, the spaces are gone and all text are in a single line.

PR #1362, opened 2023-11-14
Content block field new line

ashubawork:
- fixed a new line in the content block field (issue #737)


includes/admin/core/class-admin-builder.php
```diff
@@ -655,6 +655,7 @@ public function update_field() {
 						}
 					} elseif ( false !== strpos( $key, 'um_editor' ) ) {
 						if ( 'block' === $array['post']['_type'] ) {
+							$val                          = str_replace( "\r\n\r\n", '<br>', $val );
```
nikitasinelnikov, line 658: maybe the `nl2br()` function is the better solution?
https://www.php.net/manual/ru/function.nl2br.php

PR #1366, opened 2023-11-26
Empty fields validation

yuriinalivaiko:
Skip validation for empty fields that are not required.

The fix helps to avoid error messages like _"Please enter a valid %s ..."_ if the not-required field is empty. These error messages block profile updating.

How to reproduce an issue:

image 1

image 2


calumallison:
Task linked: [CU-86cu47cah Skip validation for empty fields that are not required](https://app.clickup.com/t/86cu47cah)

Issue #1404, opened 2023-12-29
2.8.1 added a gap under the footer of UM pages


reedock:
**Expected behavior**
A small gap under the footer of UM pages is added after latest update. 

**Isolating the problem (mark completed items with an [x]):**
- [x] I have deactivated other plugins and confirmed this bug occurs when only Ultimate Member plugin is active.
- [x] This bug happens with a default WordPress theme active, or [UM Theme](https://ultimatemember.com/theme/).
- [x] I can reproduce this bug consistently using the steps above.

**Describe the bug**
The last update 2.8.1 added a small gap under the footer of UM pages. Other pages are not having this issue. The gap is showing in the following pages: (Account, profile, directory, login, registration page etc.)

**To Reproduce**
1. Update from version 2.7.0 to 2.8.1
2. Look at UM pages, a gap is added under the footer.
3. Reverting to previous version remove the gap under the footer.

**Screenshots**
https://snipboard.io/EN2nPU.jpg

**Do you use UM extensions?**
No

**WordPress Environment**
UM 2.8.1 
WP 6.4.2 
Hello Theme
Tried with no plugins and default theme 2020 same problem with the gap.

Problem reported here without answer from support..
[Link of Wordpress topic](https://wordpress.org/support/topic/2-8-1-added-a-gap-under-the-footer)

**Temporary solution:**
Following css code added to theme:
body.um:not(.um-admin) {
margin-bottom: 0 !important;
}

alex-rosenberg35:
I would just like to add my 2 cents that I also discovered this and reported it [here](https://wordpress.org/support/topic/margin-at-bottom-of-page-and-messy-css/#post-17310663) before finding that you beat me to it.

PR #1406, opened 2024-01-01
Body styles (CU-86cu8khqt)

yuriinalivaiko:
The `um` class has been removed from the body classes to fix page styling.


calumallison:
Task linked: [CU-86cu8khqt Issue #1404: Body styles](https://app.clickup.com/t/86cu8khqt)

Issue #1407, opened 2024-01-02
Filtering by User Registration date in Members Directory


MissVeronica:
**Expected behavior**
Date selections in the Date picker 

**Describe the bug**
Date picker shows  only 1st Jan 1970

**Quick fix**
`.../plugins/ultimate-member/includes/core/class-member-directory.php` lines 820/821

```
if ( $range ) {
	if ( $filter == 'user_registered' ) {
		$min = $range[0];
		$max = $range[1];
	} else {
		$min = strtotime( $range[0] );
		$max = strtotime( $range[1] );
	}
	?>
```
**UM Support Forum  issue**
https://wordpress.org/support/topic/bug-admin-filtering-by-user-registration-date/

PR #1414, opened 2024-01-04
Member directory datepick

ashubawork:
- fix the datepicker on the member directory page (issue #1407)
- fix CSS for the datepicker in the admin area

### DIFF
--- a/assets/libs/pickadate/default.date.css
+++ b/assets/libs/pickadate/default.date.css
@@ -53,7 +53,8 @@
   width: 35%;
 }
 .um .picker__select--year {
-  width: 22.5%;
+  width: 22.5% !important;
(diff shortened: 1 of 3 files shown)


MissVeronica:
This fix will only work for User Registration date and now the filter "last_login" will fail because it's format is 2024-01-05 15:09:37

To cover all cases use
```
						if ( is_numeric( $range[0] ) ) {
							$min = $range[0];
							$max = $range[1];
						} else {
							$min = strtotime( $range[0] );
							$max = strtotime( $range[1] );
						}
```

nikitasinelnikov:
Proper PR:
https://github.com/ultimatemember/ultimatemember/pull/1425

Issue #1374, opened 2023-12-06
Welcome email with password reset link


ashubawork:
**Expected behavior**
If the registration form does not have a password field, the user should receive an email with a link to set a password.

**Describe the bug**
After registering a user without the “password” field, he receives an email with a login link, but cannot log in because he does not know his password.

**To Reproduce**
1. Go to Ultimate Members - Forms - edit Register Form - delete all fields except username/email
2. Try to register a new user
3. Check the "welcome email"
4. There is a link to the login page but without password we can't log in.

PR #1428, opened 2024-01-15
Add set password URL if user is registered without password

nikitasinelnikov:
- change placeholders for set password


nikitasinelnikov:
Task linked: [CU-86cucp72x Issue #1374: Welcome email with password reset link](https://app.clickup.com/t/86cucp72x)

Issue #1507, opened 2024-04-15
Set empty value for last_login_date while member directory sorting for users who set hidden last login date


nikitasinelnikov:
**Describe the bug**
You can see the correct last login order even for users who set the last login to hidden.

**Steps to reproduce**
1. Go to edit member directory and add last login date order
2. Set last login date hidden from the user's > Account > Privacy page
3. See that under the hood the last login date isn't empty and we don't handle the last login date privacy in the sorting

**Expected behavior**
We need to use `IF` condition in our mySQL queries to make the hidden last logins empty.

PR #1510, opened 2024-04-18
Last login order

ashubawork:
- exclude members from directory if last login is hidden (issue #1507)


nikitasinelnikov:
@ashubawork 
Please check this line in the custom meta table logic:
https://github.com/ultimatemember/ultimatemember/pull/1501/files#diff-2438a7a6506b82e50c27c0e90843bc9a34445abd2b08c993f3029c56b3c36ae6R963

We don't hide the users while sorting if they mark their last login hidden.

Please use the 'pre_user_query' hook to make the similar for WP_Users_Query.
We need to change this argument of the query: `$query->query_orderby`.

Thanks!

PR #1528, opened 2024-07-08
FEAT: parse shortcodes from post-restriction

### LINKED ISSUE #1527: Shortcodes from restricted access message aren't rendered
**Expected behavior**
As shortcodes within the block-restriction message are parsed I'd expect the same to happen for the page restriction message. Instead the page restriction content gets just outputted as a string .This makes it impossible to develop a more customized output as a message.

https://github.com/ultimatemember/ultimatemember/blob/7b8a7a7c039bde4539c07e049b19036192f1c133/includes/core/class-access.php#L764-L778

**Describe the bug**
The block-rescriction message get's parsed and therewith allow shortcodes. The page-message won't get parset and is way more limited.

**To Reproduce**
1. Create a shortcode that outputs anything
2. Create a dummy post and restrict the posts access to logged in users. Use the shortcode as output. Visit the site without beeing logged in and you'll see the shortcode as a text.
3. Create a second post and only restrict a gutenberg block. Use the shortcode as the output again. Visit the site and the shortcode will be rendered

**Expected behavior**
I'd expect a feature-consistency between the two options. As an alternative I'd expect hooks to customize the messages - at least a filter to parse the final output:
https://github.com/ultimatemember/ultimatemember/blob/7b8a7a7c039bde4539c07e049b19036192f1c133/includes/core/class-access.php#L777

Dominic-Marcelino:
fixes #1527

### DIFF
--- a/includes/core/class-access.php
+++ b/includes/core/class-access.php
@@ -774,7 +774,7 @@ if ( ! class_exists( 'um\core\Access' ) ) {
 				$content = sprintf( __( '%s', 'ultimate-member' ), $content );
 			}
 
-			return $content;
+			return do_shortcode($content);
 		}
 
 		/**


nikitasinelnikov:
Hi @Dominic-Marcelino 

I appreciate your suggestion. There is the late 'the_content' hook so that we can apply shortcodes only for the restriction message.
Also, I have added the hook for flexibility

Issue #1515, opened 2024-05-03
UM links to empty phone numbers


MissVeronica:
**Expected behavior**
A Phone field which is empty should not create a `<a href="tel:"></a>` link of the empty field

**Describe the bug**
Test for empty value before creating the link in `um-filters-fields.php`.
```
function um_profile_field_filter_hook__phone( $value, $data ) {
	$value = '<a href="' . esc_url( 'tel:' . $value ) . '" rel="nofollow" title="' . esc_attr( $data['title'] ) . '">' . esc_html( $value ) . '</a>';
	return $value;
}
```

**To Reproduce**
1. Remove the number (make blank) from an existing Phone field type `tel` of the Profile page
2. Add the Phone field to Members Directory  "Choose field(s) to display in extra user information section"
3. Look at the Members Directory and expand the User's Profile card to the extra user information section
4. You will find the Phone label and an empty tel link: 
```
<strong>Label:</strong>
<a href="tel:" rel="nofollow" title="Label"></a>
```
5. Other empty fields are not displayed in the Members Directory extra user information section but `tel` fields are when empty

**Expected behavior**
Keep empty fields empty without adding unnecessary html links

**Support Forum issue**
https://wordpress.org/support/topic/fax-number-3/

PR #1532, opened 2024-07-29
Fixed: UM links to empty phone numbers

yuriinalivaiko:
Issue: #1515 UM links to empty phone numbers

Empty field is displaying due to missing conditions. This condition fixes it;
```php
if ( empty( trim( str_replace( '+', '', $value ) ) ) ) {
  return '';
}
```
- str_replace( '+', '', $value ): Removes all + characters from $value.
- trim( ... ): Trims whitespace from the resulting string.
- empty( ... ): Checks if the resulting string is empty.

Also tel field was added and sanitized upon updating profile to other similar fields
```php
case 'tel':
```

### DIFF
--- a/includes/core/class-form.php
+++ b/includes/core/class-form.php
@@ -856,15 +856,13 @@ if ( ! class_exists( 'um\core\Form' ) ) {
 										case 'vimeo_video':
 										case 'soundcloud_track':
 										case 'spotify':
+										case 'tel':
 											$form[ $k ] = sanitize_text_field( $form[ $k ] );
 											break;
 										case 'multiselect':
 										case 'radio':
 										case 'checkbox':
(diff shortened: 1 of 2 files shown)


MissVeronica:
> so we could remove plus sign

Never remove the + sign from a telephone number as it's information for the caller's mobile
it's an international phone number and the phone is adding the caller's country's international prefix.

Example if I call the USA I will add `+1` where 1 is US country code to a local US number 
which the phone will translate to `0091` plus local number and the `009` is the international prefix.

Proper coding

```
if ( ! trim( str_replace( '+', '', $value ) ) ) {
   return '';
}
```

yuriinalivaiko:
Condition updated

MissVeronica:
Maybe this is better coding standard

```
if ( empty( trim( str_replace( '+', '', $value ) ) ) ) {
   return '';
}
```

yuriinalivaiko:
Yes, I believe that's better.

PR #1544, opened 2024-09-06
Action schedule integration

nikitasinelnikov:


includes/core/class-user.php
```diff
@@ -1667,7 +1670,7 @@ function password_changed( $user_id = null ) {
 				um_fetch_user( $user_id );
 			}
 
-			UM()->mail()->send( um_user( 'user_email' ), 'changedpw_email' );
+			UM()->maybe_action_scheduler()->enqueue_async_action( 'um_send_deleted_user_email', array( um_user( 'user_email' ), 'changedpw_email' ) );
```
nikitasinelnikov, line 1673: probably you have to use `um_mas_send_email` here too

includes/core/class-user.php
```diff
@@ -1702,17 +1705,19 @@ public function approve( $repeat = true ) {
...
 				$this->maybe_generate_password_reset_key( $userdata );
 
-				UM()->mail()->send( um_user( 'user_email' ), 'approved_email' );
+				UM()->maybe_action_scheduler()->enqueue_async_action( 'um_mas_send_email', array( $user_email, 'approved_email' ) );
 
 			} else {
 				//$userdata = get_userdata( $user_id );
 				//get_password_reset_key( $userdata );
-				UM()->mail()->send( um_user( 'user_email' ), 'welcome_email' );
+				UM()->maybe_action_scheduler()->enqueue_async_action( 'um_mas_send_email', array( $user_email, 'approved_email' ) );
```
nikitasinelnikov, line 1720: Use `welcome_email` here

includes/common/actions/class-emails.php
```diff
@@ -0,0 +1,36 @@
...
+
+	/**
+	 * Class Emails
+	 *
+	 * @package um\common\actions
+	 */
+	class Emails {
+
+		public function __construct() {
+			add_action( 'um_mas_send_email', array( $this, 'send' ), 10, 3 );
```
nikitasinelnikov, line 18: Maybe `um_send_email_action` or `um_email_sending` will be more clear name for this hook?

includes/action-scheduler/class-init.php
```diff
@@ -0,0 +1,213 @@
...
+	class Init {
+
+		protected $enabled = false;
+
+		protected $default_group = 'ultimate-member';
+
+		public function __construct() {
+			$lib_path = UM_PATH . 'includes/lib/action-scheduler/action-scheduler.php';
+			if ( ! file_exists( $lib_path ) ) {
+				wp_die( esc_html__( 'Error: Action Scheduler library is missing.', 'ultimate-member' ) );
```
nikitasinelnikov, line 24: Probably wp_die isn't a great UX, because if the user has some issues with the files he get wp_die for entire website. I think we can make the SiteHealth notification and hide the option to activate Action Scheduler.

nikitasinelnikov: Check Woocommerce action scheduler also

includes/core/class-user.php
yuriinalivaiko: These instances are fixed.

includes/core/class-user.php
yuriinalivaiko: Updated template name.

includes/action-scheduler/class-init.php
yuriinalivaiko: I’ve made a few updates. We now verify if WooCommerce is installed and activated. If it is, we then check for the presence of the Action Scheduler. If neither of these conditions is met, we check whether the Action Scheduler file exists in the Ultimate Member library directory. If everything is in order, the setting is displayed; otherwise, the setting is hidden, and we show a notification.

Issue #1546, opened 2024-09-13
Unclosed HTML tag in user description affects all the remaining elements


coiby:
**Isolating the problem (mark completed items with an [x]):**
- [ x] I have deactivated other plugins and confirmed this bug occurs when only Ultimate Member plugin is active.
- [ x] This bug happens with a default WordPress theme active.
- [ x] I can reproduce this bug consistently using the steps above.

**Describe the bug**

An unclosed HTML tag like `<s>` , `<i>`, `<b>` and etc. in user description will take effect for the whole remaining elements.

**To Reproduce**
1. Go to WP Admin -> Ultimate Member -> Settings -> Appearance then enable "HTML support for user description"
2. Put `<s>` in the user description
3. Refresh the page

**Expected behavior**
Unclosed tag in user description should not affect the remaining element.

**Screenshots**

The normal one,
![Screenshot_2024-09-13_08-11-32](https://github.com/user-attachments/assets/78536e37-5ad5-4bfd-b82f-2193dac51728)

User description with unclosed `<s>`,
![Screenshot_2024-09-13_08-10-53](https://github.com/user-attachments/assets/fdfe3869-4370-4c1a-bbf8-32bae9546e3e)

**Do you use UM extensions?**
No

**WordPress Environment**
<details>
```
`
### wp-core ###

version: 6.6.2
site_language: en_US
user_language: en_US
timezone: +00:00
permalink: undefined
https_status: false
multisite: false
user_registration: 0
blog_public: 1
default_comment_status: open
environment_type: production
user_count: 1
dotorg_communication: cURL error 5: Could not resolve proxy: squid.apac.redhat.com

### wp-paths-sizes ###

wordpress_path: /var/www/html
wordpress_size: 55.38 MB (58065519 bytes)
uploads_path: /var/www/html/wp-content/uploads
uploads_size: 10.53 MB (11042937 bytes)
themes_path: /var/www/html/wp-content/themes
themes_size: 11.92 MB (12498415 bytes)
plugins_path: /var/www/html/wp-content/plugins
plugins_size: 12.93 MB (13562080 bytes)
fonts_path: /var/www/html/wp-content/uploads/fonts
fonts_size: directory not found
database_size: 832.00 KB (851968 bytes)
total_size: 91.57 MB (96020919 bytes)

### wp-active-theme ###

name: Twenty Twenty-Four (twentytwentyfour)
version: 1.2
author: the WordPress team
author_website: https://wordpress.org
[...]

PR #1563, opened 2024-10-10
Fix HTML formatted textarea

nikitasinelnikov:
Added PHP tidy extension using to check the proper HTML value


nikitasinelnikov:
Fix for the issue:
https://github.com/ultimatemember/ultimatemember/issues/1546

PR #1619, opened 2025-01-08
Action Listeners locked to manage_options

yuriinalivaiko:
Fixed capabilities in the `actions_listener` method. Users who have the `edit_users` capability are able to change user statuses.


yuriinalivaiko:
Updated to 2.9.3-alpha

PR #1640, opened 2025-01-27
New restrictions v3

ashubawork:
- update new restrictions v3


nikitasinelnikov:
The newest branch instead of this
https://github.com/ultimatemember/ultimatemember/pull/1637

PR #1647, opened 2025-02-04
Password set in Dashboard can't be used for UM login or vice versa (CU-86cxwy7ww)

yuriinalivaiko:
Password sanitize has been changed to the standard WordPress one - no `wp_unslash`, only `trim`.


calumallison:
Task linked: [CU-86cxwy7ww Issue #1621: Password sanitize](https://app.clickup.com/t/86cxwy7ww)

PR #1653, opened 2025-02-21
Slow User page loading

yuriinalivaiko:
Fixed `set_default_user_status` function which was causing _wp-admin > Users_ page loading to be very slow if there were imported users without usermeta `account_status`.


nikitasinelnikov:
Closed by the better solution here:
https://github.com/ultimatemember/ultimatemember/pull/1670

PR #1654, opened 2025-02-28
Security path due to CVE ID: CVE-2025-1702

nikitasinelnikov:


includes/core/class-member-directory.php
```diff
@@ -1708,6 +1710,7 @@ protected function prepare_search( $search ) {
 			$regexp_map = array(
 				'/select(.*?)from/im',
 				'/select(.*?)sleep/im',
+				"/sleep\(\d+\)/im", // avoid any sleep injections
```
sjinks, line 1713: > The duration may have a fractional part.
>
> -- https://dev.mysql.com/doc/refman/8.4/en/miscellaneous-functions.html#function_sleep

Also, `SLEEP( 10 )`, `SLEEP('10')`, `SLEEP( ' 10 ' )`, etc.

There is also [`BENCHMARK()`](https://dev.mysql.com/doc/refman/8.4/en/select-benchmarking.html) function :-)

nikitasinelnikov: Hi @sjinks 

Thanks for your suggestions!
I'll apply them for the next release.

Best Regards

nikitasinelnikov: https://github.com/ultimatemember/ultimatemember/pull/1659

PR #1655, opened 2025-02-28
Updated jQuery Upload File Plugin to 4.0.11

nikitasinelnikov:


nikitasinelnikov:
It's not required with new UI and using WP native plupload JS.

PR #1659, opened 2025-03-06
Enhance SQL injection protection in regex patterns

nikitasinelnikov:
Updated regular expressions to strengthen protection against sleep and benchmark injection attacks. This change improves the security of queries by handling potential edge cases more effectively.

### DIFF
--- a/includes/core/class-member-directory.php
+++ b/includes/core/class-member-directory.php
@@ -1710,7 +1710,8 @@ if ( ! class_exists( 'um\core\Member_Directory' ) ) {
 			$regexp_map = array(
 				'/select(.*?)from/im',
 				'/select(.*?)sleep/im',
-				"/sleep\(\d+\)/im", // avoid any sleep injections
+				"/sleep\(\s*[\"'`]?\s*\d+\s*[\"'`]?\s*\)/im", // avoid any sleep injections
+				"/benchmark\(\s*[\"'`]?\s*\d+\s*[\"'`]?\s*,.*\)/im", // avoid any benchmark injections
 				'/select(.*?)database/im',
 				'/select(.*?)where/im',
 				'/update(.*?)set/im',


includes/core/class-member-directory.php
```diff
@@ -1710,7 +1710,8 @@ protected function prepare_search( $search ) {
 			$regexp_map = array(
 				'/select(.*?)from/im',
 				'/select(.*?)sleep/im',
-				"/sleep\(\d+\)/im", // avoid any sleep injections
+				"/sleep\(\s*[\"'`]?\s*\d+\s*[\"'`]?\s*\)/im", // avoid any sleep injections
```
sjinks, line 1713: What do you think about

```suggestion
				"/sleep\([^)]+)/im", // avoid any sleep injections
```

nikitasinelnikov: Thanks @sjinks, it looks easier and works. It will also catch no-standard syntax inside sleep.

sjinks:
@nikitasinelnikov I have a naive question (I am not very familiar with the codebase): why don't just

```php
global $wpdb;
$search = trim( sanitize_text_field( wp_unslash( $search ) ) );
return $wpdb->escape( '%s', $search );
```

This will return `$search` as a properly escaped string enclosed in quotes.

Then, if you don't need quotes, like in `general_search()` for `meta_query`, you just `substr( $search, 1, -1 )`. In other places, you use the quoted string, ensuring it is used as a string and not as a possible raw SQL.

nikitasinelnikov:
Hi @sjinks 

`$wpdb->escape` has been deprecated since 3.6.0.
As you can see, we use the alternative `esc_sql()` function below, but it doesn't handle a string for removing insecure SLEEP or BENCHMARK MySQL functions.

https://github.com/ultimatemember/ultimatemember/blob/fix/1654_additional/includes/core/class-member-directory.php#L1729

Thanks for your suggestions!

nikitasinelnikov:
Hi @sjinks 

I have updated our handlers in this PR:
https://github.com/ultimatemember/ultimatemember/pull/1663

I avoid using regexp for direct replacements in MySQL queries.

Best Regards!

PR #1744, opened 2025-09-15
Dev/champ new UI

champsupertramp:
Using this method:

```
echo UM()->frontend()::layouts()::button( '', $btn_arr );
```

When a button doesn't require a label and only an icon, we need to remove or don't add the content elements so that it doesn't add extra gap in the button to center the icon:

Before:
Screenshot 2025-09-16 at 3 45 25 AM

After applying the code in this PR:
Screenshot 2025-09-16 at 3 46 40 AM


nikitasinelnikov:
Hi @champsupertramp
It is possible to put the icon HTML directly in the content argument and use 'icon_position' => 'content' without the 'icon' attribute.

Please take a look at this reference and try the same in your code:
https://github.com/ultimatemember/bbpress/blob/302b1ecc8cf9bc7f9ea2f0c103f35fee5eacb80c/templates/v3/favorites.php#L60-L84

nikitasinelnikov:
Hi @champsupertramp 

See the commit https://github.com/ultimatemember/ultimatemember/commit/a997284eb0f1cc41da1f37cc43973166e97b24ef

There is possible to extend the icons list via filter hook:

```
add_filter( 'um_config_get', array( &$this, 'extends_icons' ), 10, 2 );
/**
* Extends icons in the config.
*
* @param mixed  $config The config value to be extended.
* @param string $key    The key of the config value.
*
* @return mixed The updated config value if key is 'svg_icons', otherwise the original config value.
*/
public function extends_icons( $config, $key ) {
  if ( 'svg_icons' !== $key ) {
	  return $config;
  }
  $config_new = array(
	  'shopping-cart' => '<svg  xmlns="http://www.w3.org/2000/svg"  width="24"  height="24"  viewBox="0 0 24 24"  fill="currentColor"  class="icon icon-tabler icons-tabler-filled icon-tabler-shopping-cart"><path stroke="none" d="M0 0h24v24H0z" fill="none"/><path d="M6 2a1 1 0 0 1 .993 .883l.007 .117v1.068l13.071 .935a1 1 0 0 1 .929 1.024l-.01 .114l-1 7a1 1 0 0 1 -.877 .853l-.113 .006h-12v2h10a3 3 0 1 1 -2.995 3.176l-.005 -.176l.005 -.176c.017 -.288 .074 -.564 .166 -.824h-5.342a3 3 0 1 1 -5.824 1.176l-.005 -.176l.005 -.176a3.002 3.002 0 0 1 1.995 -2.654v-12.17h-1a1 1 0 0 1 -.993 -.883l-.007 -.117a1 1 0 0 1 .883 -.993l.117 -.007h2zm0 16a1 1 0 1 0 0 2a1 1 0 0 0 0 -2zm11 0a1 1 0 1 0 0 2a1 1 0 0 0 0 -2z" /></svg>',
  );
  return array_merge( $config, $config_new );
}
```

Thanks for your suggestions

PR #1776, opened 2026-01-21
Use a hook to activate an account

yuriinalivaiko:
Use the `um_after_email_confirmation` hook to approve the user after the activate account link is verified.

This may be helpful for website developers who need to customize the account activation logic.

These changes were suggested by David in ticket 108875.


nikitasinelnikov:
Please use new hook for that:
`um_approve_user_on_email_confirmation`

PR #1803, opened 2026-03-31
Guests tokens

ashubawork:
- add the schedule action to delete guests tokes from the DB


includes/common/actions/class-guests.php
```diff
@@ -0,0 +1,128 @@
...
+
+			$tokens_table   = "{$wpdb->prefix}um_guest_tokens";
+			$attempts_table = "{$wpdb->prefix}um_guest_download_attempts";
+
+			/**
+			 * How many days guest tokens should live.
+			 *
+			 * @param int $days
+			 */
+			$tokens_ttl_days = absint( apply_filters( 'um_guest_tokens_ttl_days', 0 ) );
```
nikitasinelnikov: @ashubawork can be there 1 day by default?

includes/common/actions/class-guests.php
```diff
@@ -0,0 +1,128 @@
...
+					"SELECT token
+					FROM {$tokens_table}
+					WHERE created_at < NOW() - INTERVAL %d DAY",
+					$tokens_ttl_days
+				)
+			);
+
+			// Remove temp folders for collected tokens.
+			foreach ( $expired_guest_tokens as $token ) {
+				$token = sanitize_file_name( $token );
```
nikitasinelnikov: I'm not sure we need sanitize_file_name here, because the token has already been stored in DB.

includes/common/actions/class-guests.php
```diff
@@ -0,0 +1,128 @@
...
+
+				if ( empty( $token ) ) {
+					continue;
+				}
+
+				$folder_path = trailingslashit( $temp_folder ) . $token;
```
nikitasinelnikov: wp_normalize_path for entire  $temp_folder . $token. I'm not sure about trailingslashit. Temp folder always has the end slash.

includes/common/actions/class-guests.php
```diff
@@ -0,0 +1,128 @@
...
+			$wpdb->query(
+				$wpdb->prepare(
+					"DELETE FROM {$attempts_table}
+					WHERE request_time < NOW() - INTERVAL %d DAY",
+					$attempts_ttl_days
+				)
+			);
+
+			// Reset auto-increment value (if $tokens_ttl_days and $attempts_ttl_days = 0)
+			if ( 0 === $attempts_ttl_days && 0 === $attempts_ttl_days ) {
```
nikitasinelnikov: I think it's unnecessary condition. But the lines

$wpdb->query("ALTER TABLE {$tokens_table} AUTO_INCREMENT = 1");
$wpdb->query("ALTER TABLE {$attempts_table} AUTO_INCREMENT = 1");

below without this condition will be enough

nikitasinelnikov (changes requested):

includes/common/actions/class-guests.php
ashubawork: not working without trailingslashit

nikitasinelnikov: @ashubawork, sorry I was wrong.

$folder_path = wp_normalize_path( $temp_folder . DIRECTORY_SEPARATOR . $token );

nikitasinelnikov: $temp_folder is without a slash at the end

includes/common/actions/class-guests.php
```diff
@@ -0,0 +1,121 @@
...
+
+			/**
+			 * How many days download attempts should live.
+			 *
+			 * @param int $days
+			 */
+			$attempts_ttl_days = absint( apply_filters( 'um_guest_download_attempts_ttl_days', 1 ) );
+
+			if ( UM()->is_new_ui() ) {
+				$temp_folder = UM()->common()->filesystem()->get_tempdir();
```
nikitasinelnikov, line 76: Guest tokens are related to the new UI only, so I think we can leave only

$temp_folder = UM()->common()->filesystem()->get_tempdir();

PR #1821, opened 2026-06-29
oEmbed field

yuriinalivaiko:
Fixed incorrect sizes of the `iframe` element in the oEmbed field.

Issue screenshot:
image

Fixed screenshot:
image


assets/js/um-functions.js
```diff
@@ -250,8 +250,19 @@ function um_responsive(){
 			jQuery('.um-account-side li a.current').trigger('click');
 		}
 
-		jQuery(this).css('opacity',1);
+		// set iframe height and width.
+		const $iframes = jQuery(this).find('.um-field-value iframe[height][width]');
```
nikitasinelnikov, line 254: Hi @yuriinalivaiko 
Is it possible to set the CSS hard-coded "width:100% + height:auto" properties for the oembed blocks to avoid the JS handlers?

Or is this solution created for non-full-width oembed blocks?

yuriinalivaiko: Hi @nikitasinelnikov 
I tried to solve this problem using CSS first. But the `height:auto` or `height:100%` does not act well in `iframe`.
See this example https://ibb.co/NnxDXxY6

PR #1823, opened 2026-06-30
Embed local resources

yuriinalivaiko:
The "oEmbed" field has been updated to support embedding local resources, such as a local video from the website Media Library.

### DIFF
--- a/includes/core/class-fields.php
+++ b/includes/core/class-fields.php
@@ -4604,12 +4604,15 @@ if ( ! class_exists( 'um\core\Fields' ) ) {
 					if ( isset( $data['label'] ) || ! empty( $data['icon'] ) ) {
 						$output .= $this->field_label( $data['label'], $key, $data );
 					}
-					$response = wp_oembed_get( $_field_value );
-					if ( empty( $response ) ) {
-						$response = $_field_value;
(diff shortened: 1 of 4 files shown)


includes/core/class-fields.php
```diff
@@ -4604,12 +4604,15 @@ public function view_field( $key, $data, $rule = false ) {
 					if ( isset( $data['label'] ) || ! empty( $data['icon'] ) ) {
 						$output .= $this->field_label( $data['label'], $key, $data );
 					}
-					$response = wp_oembed_get( $_field_value );
-					if ( empty( $response ) ) {
-						$response = $_field_value;
+					if ( ! empty( esc_url( $_field_value ) ) ) { // This variable can contain a URL or HTML code.
```
nikitasinelnikov, line 4607: @yuriinalivaiko I think this double fallback with `wp_oembed_get()` can be removed. The function `um_profile_field_filter_hook__oembed()` prepares the field_value using `wp_oembed_get()` and other functions

nikitasinelnikov: I'll make these changes after the merge. Thanks

yuriinalivaiko: Thank you. I noticed the `wp_oembed_get()`  function was being used twice, but I didn't dare delete it. I just added a condition to the second use.

PR #1826, opened 2026-07-03
fix(shortcodes): prevent infinite recursion when GDPR policy page contains UM shortcode

### LINKED ISSUE #1809: Registration form infinite loop - gdpr-register.php calls the_content recursively causing PHP fatal error
**Ultimate Member Version:** 2.11.4
**WordPress Version:** 6.x
**Theme:** Hello Elementor
**Plugins:** Elementor Pro 4.0.2

**Describe the bug**
The registration form creates an infinite loop on the registration page. The page never finishes loading and ends with a PHP fatal error: Maximum call stack size reached.

**Root cause found in debug.log:**
`ultimate-member/templates/gdpr-register.php` line 28 calls `apply_filters('the_content', '[ultimatemember...')` which re-triggers the ultimatemember shortcode → loads register.php → calls gdpr-register.php again → infinite loop.

**To Reproduce**
1. Assign a page as registration page in UM Settings → Pages
2. Add shortcode [ultimatemember form_id="24"] to the page
3. Visit the page → infinite loop until PHP fatal error

**Expected behavior**
Registration form loads once without loop.

**Debug log excerpt:**
PHP Fatal error: Uncaught Error: Maximum call stack size reached. Infinite recursion?
#22 gdpr-register.php(28): apply_filters('the_content', '[ultimatemember...', 30)
#23 class-gdpr.php(39): require('gdpr-register.php')
#27 register.php(35): do_action('um_after_form_f...')
#28 class-shortcodes.php(362): include('register.php')
→ repeats infinitely

faisalahammad:
## Summary

Add recursion guard in `Shortcodes::ultimatemember()` to prevent infinite render loop when the GDPR privacy policy page content contains a `[ultimatemember]` shortcode.

Fixes #1809

### `includes/core/class-shortcodes.php`
**Before:**
```php
// No guard — if the_content filter re-triggers [ultimatemember] shortcode mid-render, infinite loop
return $this->load( $args );
```
**After:**
```php
static $rendering = array();
if ( in_array( $args['form_id'], $rendering, true ) ) {
    return '';
}
$rendering[] = $args['form_id'];
try {
    // ... existing singleton guard and $this->load($args)
    return $this->load( $args );
} finally {
    array_pop( $rendering );
}
```
**Why:** `gdpr-register.php` runs `apply_filters('the_content', $content)` on the privacy policy page content. If that page contains a `[ultimatemember]` shortcode, `do_shortcode` fires inside the content filter, re-rendering the form, re-firing the GDPR hook, and looping infinitely. The static `$rendering` stack detects nested re-entry with the same `form_id` and bails. Sequential renders (same form twice on a page) are unaffected since the stack is empty between top-level calls.

## Testing

**Test 1: Recursion on misconfigured privacy policy page**
1. Create a registration form (note form_id, e.g. 24)
2. Enable GDPR in form settings, set "Privacy Policy Page" to a page containing `[ultimatemember form_id="24"]`
3. Visit the registration page

Result: Before fix, PHP fatal "Maximum call stack size reached". After fix, page loads normally, GDPR checkbox renders.

**Test 2: Sequential same-form rendering still works**
1. Place two `[ultimatemember form_id="24"]` shortcodes on the same page (not nested, sequential)
2. Visit page

Result: Both forms render correctly. The recursion guard does not interfere with sequential renders since the stack is empty between calls.

## Screenshots
SCR-20260703-rkws
-
SCR-20260703-rkur


nikitasinelnikov:
Added to https://github.com/ultimatemember/ultimatemember/pull/1837
Using the page select and render additional conditions.

PR #1834, opened 2026-07-06
Store api_key settings as wp-config.php constants

champsupertramp:
# Store `api_key` settings as `wp-config.php` constants

## Summary
Secret values entered into `api_key` settings fields (e.g. um-stripe's publishable/secret keys) are now stored as PHP constants in `wp-config.php` instead of in the `um_options` database row. Reads resolve transparently, so consuming extensions need no changes.

## Motivation
API keys are secrets and shouldn't sit in the database. Storing them as `wp-config.php` constants keeps them out of the DB (and out of DB backups/exports), consistent with the reveal/hide direction of the `api_key` field type.

## How it works
- **Constant naming:** `UM_OPTION_<UPPERCASE_ID>` (e.g. `stripe_test_secret_key` → `UM_OPTION_STRIPE_TEST_SECRET_KEY`). Overridable via the `um_option_constant_name` filter.
- **Scoped to `api_key` fields only:** a registry of constant-backed option ids is derived from the settings structure and persisted to the `um_api_key_option_ids` option (so it's available on the frontend, where the settings structure isn't built). Extend via the `um_api_key_option_ids` filter. Non-`api_key` options are never constant-searched.
- **Safe writes:** uses the `wp-cli/wp-config-transformer` library to add/update/remove constants — updates in place (never duplicates a `define()`).

## Changes
- **`composer.json`** — add `wp-cli/wp-config-transformer` to `require`; add a `wp-config-transformer` script (run on `post-install`/`post-update`) that copies the single class file into `includes/lib/wp-config-transformer/`, mirroring the Action Scheduler / php-scoper-excludes bundling convention.
- **`.gitignore`** — ignore `includes/lib/wp-config-transformer/` (build-generated, shipped in the release).
- **`includes/admin/core/class-wp-config.php`** *(new)* — thin wrapper over `WPConfigTransformer`: resolves the `wp-config.php` path (with the standard non-subdir fallback), lazy-loads the bundled lib, and exposes `set_constant()` / `remove_constant()` with try/catch.
- **`includes/core/class-options.php`** — new `get_constant_name()`, `is_constant_backed()`, `get_constant_backed_ids()`, `set_constant_backed_ids()`. `get()` now returns a defined constant's value (for constant-backed ids) ahead of the DB, falling back to the DB when undefined.
- **`includes/admin/core/class-admin-settings.php`** — `get_api_key_field_ids()` walks the settings structure for `type => api_key` fields; `save_api_key_constants()` (on `um_change_settings_before_save`) writes/removes the constant, strips the value from the saved settings, and deletes any legacy DB copy so the secret is never persisted to `um_options`; `init_variables()` keeps the constant-backed registry in sync.
- **`includes/admin/core/class-admin-notices.php`** — three notices: bundled library missing, wp-config write failure (with the exact `define(...)` lines to add manually), and a migration prompt when keys still live in the DB (links to the relevant settings section to re-save).

## Behavior / UX
- Reading a key prefers the constant; empty submission removes the constant.
- **Migration:** existing keys stored in the DB keep working (read falls back to the DB) and an admin notice prompts a re-save to move them into `wp-config.php`.
- **Write failure** (e.g. `wp-config.php` not writable): no DB fallback — an admin notice shows the exact constants to add manually, and the legacy DB value is preserved so the site keeps working.

## Notable fixes included
- Re-saving an unchanged key no longer reports a false "could not be written" error (WPConfigTransformer's `save()` returns `false` for an unchanged file; treated as success).
- After writing a constant, the OPcache entry for `wp-config.php` is invalidated so the new value is visible on the very next request (the post-save redirect) instead of requiring a second reload.

## Backward compatibility
- No changes required in consuming extensions (um-stripe untouched) — `UM()->options()->get('stripe_*_key')` transparently resolves to the constant.
- Only `api_key` fields are affected; all other options behave exactly as before.

## Packaging note
`includes/lib/wp-config-transformer/` is gitignored by design; the release pipeline must run `composer install` (the `post-install`/`post-update` script copies the lib into place).

## Testing
- Save/update/remove keys → correct `define()` in `wp-config.php` (exactly one line, no duplicates); no `stripe_*_key` entry remains in `um_options`.
- Constant read wins; DB fallback works when no constant is defined; unrelated options and `site_name`/`admin_email` special-cases unaffected.
- Write-failure path returns `false` on a read-only `wp-config.php` and surfaces the manual-instructions notice.
- Re-save with an unchanged value succeeds silently; freshly saved keys display after the redirect without a second reload.


includes/admin/class-wp-config.php
```diff
@@ -0,0 +1,159 @@
+<?php
+namespace um\admin;
```
nikitasinelnikov, line 2: Hi @champsupertramp 

All new admin classes in UM core it's better to place into um\admin namespace. All um\admin\core namespace classes will be moved to um\admin soon. It's just a future todo.

composer.json
nikitasinelnikov, line 51: composer install command uses the installer-paths. I have applied the same for wp-config-transformer. The necessary package is installed to the releases or per-build branch.

includes/admin/class-admin.php
```diff
@@ -2141,6 +2141,18 @@ public function users_columns() {
 			return UM()->classes['um\admin\users_columns'];
 		}
 
+		/**
+		 * @since 2.13.0
+		 *
+		 * @return WP_Config
+		 */
+		public function wp_config() {
```
nikitasinelnikov, line 2149: Added a singleton helper for the WP_Config class. So `new WP_Config()` is transformed to `$wp_config = UM()->admin()->wp_config();`

PR #1843, opened 2026-07-20
fix(admin): scope um_admin_global enqueue to UM-owned screens

### LINKED ISSUE #1842: Ultimate Member causes site-wide rest_cookie_invalid_nonce on all authenticated REST requests
Ultimate Member (latest version) invalidates REST nonces site-wide on my install. Every authenticated REST request fails with {"code":"rest_cookie_invalid_nonce","message":"Cookie check failed"} — across unrelated plugins simultaneously (Elementor, Real Media Library, admin-ajax.php), not just UM's own endpoints. WordPress then shows its "Your session has expired. Please log in to continue" modal, and sometimes redirects to wp-login.php?reauth=1. It occurs during editing and spontaneously while idle.
Isolation performed:
Deactivated all plugins, then reactivated one at a time (fresh login, cleared cookies, console filtered to nonce errors, tested 15+ min each). With every plugin active EXCEPT Ultimate Member: no errors, worked normally. Reactivated Ultimate Member: immediately prompted to log in again, console filled with rest_cookie_invalid_nonce. Fully reproducible — UM off = works, UM on = fails.
Already ruled out:

Server/page cache — host purged all server-side + LiteSpeed cache, no change. LiteSpeed logged-in caching and heartbeat control both OFF.
Host infrastructure — Hostinger checked their live request logs at a failure timestamp and confirmed the rejections come from WordPress's own cookie/nonce authentication, not their cache/firewall/rate-limiting. IP not blocked.
Hardcoded COOKIEHASH in wp-config — tested by commenting out with cleared cookies + fresh login; errors persisted. Restored.
Site URL mismatch — WP and Site URLs identical, https, no [www](http://www/). Site never migrated.
Fresh auth salts and DB password — rotated recently; errors persist.
UM's own "Enable Rate Limiting" setting — unchecked and saved; errors persist.
"Remember Me" does not extend the session.

Also observed: JS error on every editor load — Uncaught TypeError: Cannot read properties of undefined (reading 'hasClass') in wp-auth-check.min.js via heartbeat.min.js. Possibly related, as heartbeat refreshes the nonce.
Environment: WordPress 6.9.x, Ultimate Member (latest), Elementor Pro, ACF Pro, Real Media Library Pro, LiteSpeed Cache, SearchWP, FluentCommunity/CRM/SMTP/Auth, Code Snippets, Frontend Admin. Hosted on Hostinger. Chrome/Windows; also reproduces on iOS Safari.

faisalahammad:
## Summary
UM was enqueuing the global admin script, CSS, and a localized um-admin-nonce on every wp-admin page, including unrelated screens like the Elementor editor, Real Media Library, and ACF. That interfered with wp-auth-check and heartbeat load order in WP 6.9 and caused the wp-auth-check TypeError that broke the nonce-refresh path. The breakage surfaced site-wide as `rest_cookie_invalid_nonce` errors and a "Your session has expired" modal that affected unrelated plugins.

Scope the enqueue to `is_own_screen()` with a new `um_enqueue_global_admin_scripts` filter for opt-in. The script still loads on every UM admin screen and the dismiss handler continues to work.

Fixes #1842

### includes/admin/class-enqueue.php
**Before:**
```php
public function admin_enqueue_scripts( $hook ) {
    $suffix  = self::get_suffix();
    $js_url  = self::get_url( 'js' );
    $css_url = self::get_url( 'css' );

    $this->load_global_scripts();

    if ( UM()->admin()->screen()->is_own_screen() ) {
```
**After:**
```php
public function admin_enqueue_scripts( $hook ) {
    $suffix  = self::get_suffix();
    $js_url  = self::get_url( 'js' );
    $css_url = self::get_url( 'css' );

    // Load the global admin script only on UM-owned screens.
    // Enqueueing it on every wp-admin page interferes with wp-auth-check/heartbeat
    // load order and caused the site-wide rest_cookie_invalid_nonce regression (GitHub issue #1842).
    $enqueue_global = UM()->admin()->screen()->is_own_screen();
    /**
     * Filters whether the global UM admin script and styles are enqueued.
     *
     * @since 2.12.2
     * @hook um_enqueue_global_admin_scripts
     *
     * @param {bool}   $enqueue_global Whether to enqueue global admin assets. Default is the result of `is_own_screen()`.
     * @param {string} $hook           Current wp-admin screen hook.
     *
     * @return {bool} True to enqueue, false to skip.
     */
    $enqueue_global = apply_filters( 'um_enqueue_global_admin_scripts', $enqueue_global, $hook );
    if ( $enqueue_global ) {
        $this->load_global_scripts();
    }

    if ( UM()->admin()->screen()->is_own_screen() ) {
```
**Why:** Removes the global JS+CSS+localized-nonce injection on every wp-admin page. The global script is only needed for the UM admin notice dismiss handler, which only renders on UM-owned screens. The new filter gives extension authors a documented opt-in path for edge cases.

A short note was also added to the `load_global_scripts()` PHPDoc to flag the caller-side gate so future readers do not reintroduce the regression by re-calling it site-wide from elsewhere.

## Testing

**Test 1: Site-wide rest_cookie_invalid_nonce on third-party admin pages (issue #1842)**
1. WP 6.9.x + UM 2.12.1 + Elementor Pro + Real Media Library Pro + ACF Pro + LiteSpeed Cache
2. Log in as admin in a fresh browser, open DevTools console and Network
3. Open Elementor editor, open a page, save, close; repeat
4. Open Real Media Library, perform an action
5. Open ACF field group, save
6. Idle for 5 minutes with Network recording

Result: no `wp-auth-check.min.js` TypeError in console; REST calls to `/wp-json/...` return 200; heartbeat ticks succeed; "Your session has expired" modal does not appear.

**Test 2: UM admin still works on UM-owned screens**
1. Go to wp-admin -> Users -> All Users
2. Confirm any UM admin notice dismiss X is clickable and persists dismissal
3. Go to wp-admin -> Ultimate Member -> Settings, repeat
4. Edit a UM Form, confirm the builder loads and the upgrade page (`um_upgrade`) AJAX still works
5. View a user profile at `/user/<slug>/` and confirm UM fields render

Result: notices dismiss, builder loads, AJAX upgrade flow works, profile fields render.

**Test 3: Filter opt-in for custom UM notice surfaces**
1. Add to a mu-plugin:
   ```php
   add_filter( 'um_enqueue_global_admin_scripts', '__return_true' );
   ```
2. Visit any wp-admin page (e.g. `plugins.php`) where a custom UM notice renders
3. Confirm the dismiss X is now clickable

Result: filter forces the script back on.

## Edge case
If a UM admin notice happens to render on the wp-admin dashboard (`index.php`), the dismiss button will be inert because `is_own_screen()` returns false there. The `um_enqueue_global_admin_scripts` filter is the documented opt-in for that case.


nikitasinelnikov:
Hi @faisalahammad 

Please check this PR. We need a global JS for global admin notices
https://github.com/ultimatemember/ultimatemember/pull/1845

So I just removed the localize scripts and added ignore global JS/CSS assets by the $hook (screen of the enqueue scripts variable) to avoid conflicts via 3rd-party integrations.

Best Regards

nikitasinelnikov:
This PR can be closed because it was merged to 2.13.0 version.

PR #1845, opened 2026-07-21
Add nonce checks and filters for admin notice dismissal process

### LINKED ISSUE #1842: Ultimate Member causes site-wide rest_cookie_invalid_nonce on all authenticated REST requests

nikitasinelnikov:
Refactored admin notice handling to enhance security and flexibility by introducing nonce verification for both AJAX and manual dismissals. Added new filters to customize global script loading and notice display logic.


ashubawork (commented):
Looks good to me.
Compared with #1843, this seems like a better fix: instead of disabling `um_admin_global` outside UM screens, it removes the global `um_admin_scripts.nonce` dependency and uses per-notice nonces for admin notice dismissal. I don’t see any blockers.

michaeldavisdcpersonal:
Reporter here (#1842). Thanks again for turning this around quickly. I see #1845 has ashubawork's review and has been sitting on `development/2.9.x` for a couple of weeks now — is there a milestone or target release it's slated for? Happy to help however's useful.

If it helps move it along, I can validate `fix/avoid-global-admin-nonce` on the exact environment that triggered the regression — Hostinger + LiteSpeed Cache + WP 6.9.x + Elementor Pro / ACF Pro / Real Media Library Pro, which I don't think is trivial to reproduce otherwise. Since the failure only surfaces after a nonce expires, I'd run it through a full nonce/heartbeat cycle (~24h) and confirm the `wp-auth-check` TypeError is gone and no `rest_cookie_invalid_nonce` 403s appear, including while idle. Just say the word.

nikitasinelnikov:
Hi @michaeldavisdcpersonal sorry for the delay. 
This patch will be released this or next week. I'm waiting for feedback about another security patch.

And I just lost this PR's review results =)

Best Regards!

michaeldavisdcpersonal:
Thank you!

michaeldavisdcpersonal:
Hi There, any sense of when this update will hit?

Thanks for your help!

PR #1847, opened 2026-07-22
fix(member-directory): add filter for computed profile field values in directory dropdown

### LINKED ISSUE #596: Filter um_profile_{$key}__filter is working on profile page but not on directory search form
## My Integration of UM plugin

1. I have created custom post type called **Hospital** and it has custom field with *meta_key* as `doctor_id`. In each **Hospital** post, I assign **ID** of all doctors *(user profile as doctor of all doctors is already created)* working in that hospital to this custom field `doctor_id`. 

2. This way I can query all the **Hospital Names** that each doctor is assigned to. The query is inside a callback function that returns an array of hospital names where a doctor is assigned.

3. I can simply list the **Hospital Names** in a doctor profile from this callback function but I want to make visitors filter doctors by **Hospital Name** in members directory page. So I needed to create a select field which I can then *enable in search*.

4. So, In *Profile Form*, I have created a multi-select field on with key as `hospital_list`. I can't create this field without adding choices. So, I used *Choices Callback* to populate *Edit Choices* field which basically populates name of all **published Hospital posts**.
 
## Things that Worked

- [x] I used following two filters to set multi-select field value using a callback function:
`um_profile_{$key}__filter` which will be `um_profile_hospital_list__filter`
`um_profile_{$key}_empty__filter` which will be `um_profile_hospital_list_empty__filter`

- [x] When I `var_dump` the *function* I mentioned in **Point 3** above and when I `var_dump` the field value using *um_user( 'hospital_list' )*, they both produce same result. That means above filter is auto-populating `hospital_list` field as expected. Even if user edits the profile and try to de-select one of the hospital, the filter won't allow to do that which is what I wanted.

## Things that didn't work

- [ ] Although the filter is setting the value, the **Hospital** search drop-down doesn't work as expected.
- [ ] When a follow **Step 1** to add/remove **ID** of doctor from a **Hospital Post**, I have to edit and update the user profile also whose **ID** has be added/removed. Only this way does the search query works.

## My Query
 - Is there something I am missing for the field to be auto-updated instead of editing and updating the profile again?

**WordPress Environment**
<details>
```
### Begin Install Info ###


--- Site Info ---

Site URL:					http://hos-list.demo
Home URL:					http://hos-list.demo
Multisite:					No

--- Hosting Provider ---

Host:						DBH: localhost, SRV: hos-list.demo

--- User Browser ---

Platform:                 Windows 
Browser Name:             Firefox  
Browser Version:          70.0 
User Agent String:        Mozilla/5.0 (Windows NT 10.0; Wi 
			  n64; x64; rv:70.0) Gecko/2010010
                          1 Firefox/70.0
                          
---- Current User Details --

Role: administrator


--- WordPress Configurations ---

Version:						5.2.2
Language:					en_US
Permalink Structure:			/%category%/%postname%/
Active Theme:				Flatsome Child 3.0
Page On Front:				home (#2)
Page For Posts:				 (#0)
ABSPATH:					/app/public/
All Posts/Pages:				10
WP Remote Post:           		wp_remote_post() works
WP_DEBUG:                 			Disabled
WP Table Prefix:          			Length: 3, Status: Acceptable
Memory Limit:   				40MB


--- UM Configurations ---

Version:						2.0.56
Upgraded From:            		2.0.54
Current URL Method:			
Cache User Profile:			No
Generate Slugs on Directories:	Yes
Rewrite Rules: 				Yes
Force UTF-8 Encoding: 		No
JS/CSS Compression: 			No
Port Forwarding in URL: 		No
Exclude CSS/JS on Home: 		No


--- UM Pages Configuration ---

User:						http://hos-list.demo/doctor/
Account:						http://hos-list.demo/account/
Members:					http://hos-list.demo/
Register:						http://hos-list.demo/login/
Login:						http://hos-list.demo/login/
Logout:						http://hos-list.demo/logout/
Password Reset:				http://hos-list.demo/reset-password/


--- UM Users Configuration ---

Default New User Role: 		0
Profile Permalink Base:		name_dash
User Display Name:			full_name
Force Name to Uppercase:		No
Redirect author to profile: 		Yes
Enable Members Directory:	Yes
Use Gravatars: 				No
Require a strong password: 	Off


--- UM Access Configuration ---

Panic Key: 								
Global Site Access:						Site accessible to Everyone
Backend Login Screen for Guests:			No
Redirect to alternative login page:			
Backend Register Screen for Guests:		No
Redirect to alternative register page:		
Access Control widget for Admins only: 		No
Enable the Reset Password Limit:			Yes
Reset Password Limit: 3Disable Reset Password Limit for Admins: No
Blacklist Words: 							5


--- UM Email Configurations ---

Mail appears from:  			Om Hospital
Mail appears from address:  	om@admin.com
Use HTML for E-mails:   		Yes
Account Welcome Email:  		Yes
Account Activation Email:   	No
Pending Review Email:   		No
Account Approved Email: 		No
Account Rejected Email: 		No
Account Deactivated Email:  	Yes
Account Deleted Email:  		Yes
Password Reset Email:   		Yes
Password Changed Email: 		Yes


--- UM Total Users ---

                All Users(9)
administrator(1)
contributor(7)
customer(1)
none(0)


--- UM Roles ---

                Administrator (administrator)
Editor (editor)
Author (author)
Contributor (contributor)
Subscriber (subscriber)
Customer (customer)
Shop manager (shop_manager)


--- UM Custom Templates ---

                File: /app/public/wp-content/themes/flatsome-child/ultimate-member/templates/login-custom.php
File: /app/public/wp-content/themes/flatsome-child/ultimate-member/templates/members.php
File: /app/public/wp-content/themes/flatsome-child/ultimate-member/templates/profile-twocol.php


--- UM Email HTML Templates ---

                N/A


--- Web Server Configurations ---

PHP Version:              			7.3.2
MySQL Version:            			5.6.34
Web Server Info:          			Apache/2.4.10 (Debian)


--- PHP Configurations ---

PHP Memory Limit:         			256M
PHP Upload Max Size:      			1000M
PHP Post Max Size:        			1000M
PHP Upload Max Filesize:  			1000M
PHP Time Limit:           			900
PHP Max Input Vars:       			3000
PHP Arg Separator:        			&
PHP Allow URL File Open:  			Yes


--- Web Server Extensions/Modules ---

DISPLAY ERRORS:           			On (1)
FSOCKOPEN:                			Your server supports fsockopen.
cURL:                     			Your server supports cURL.
SOAP Client:              			Your server has the SOAP Client enabled.
SUHOSIN:                  			Your server does not have SUHOSIN installed.
GD Library:               			PHP GD library is installed on your web server.
Mail:                     			PHP mail function exist on your web server.
Exif:				          PHP Exif library is installed on your web server.


--- Session Configurations ---

Session:                  			Disabled
Session Name:             			PHPSESSID
Cookie Path:              			/
Save Path:                			
Use Cookies:              			On
Use Only Cookies:         			On


--- WordPress Active Plugins ---

                Advanced Custom Fields PRO: 5.8.3
Contact Form 7: 5.1.4
Font Awesome: 4.0.0-rc13
Hospital Directory Listing: 1.1.0
Page scroll to id: 1.6.6
Ultimate Member: 2.0.56
WooCommerce: 3.7.0
WooCommerce Admin: 0.19.0
WooCommerce Variation Swatches and Photos: 3.0.12

### End Install Info ###
                    
```
</details>

faisalahammad:
## Summary

`um_profile_{$key}__filter` lets developers transform a field value at render time (e.g. populate a multiselect from an external post type, derive a label). Because the value is computed and never stored in `wp_usermeta`, the member-directory filter dropdown was always empty for those fields.

This adds one new opt-in filter at the single chokepoint where dropdown values are finalized, so developers can supply synthetic options without patching core.

Fixes #596

**Before** — dropdown values came only from `SELECT DISTINCT meta_value FROM $wpdb->usermeta`, with an early-return when the result was empty:

```php
// no way for a developer to supply values for fields
// whose meta never gets stored
if ( ! empty( $values_array ) && in_array( $attrs['type'], array( ... ), true ) ) {
```

**After** — one new `apply_filters()` call right after the DB result is finalized, before the existing intersection and post-filter:

```php
$values_array = apply_filters(
    "um_member_directory_filter_values_{$attrs['metakey']}",
    $values_array,
    $attrs
);

if ( ! empty( $values_array ) && in_array( $attrs['type'], array( ... ), true ) ) {
```

**Why:** minimal-blast-radius fix at the natural extension point. BC-safe — when no dev hooks the new filter, `apply_filters()` returns the same value it received, so behavior is byte-identical to current master.

The search side is intentionally unchanged. Developers wire the existing `um_query_args_{$field}__filter` (standard `WP_User_Query` path) or `um_query_args_{$field}__filter_meta` (custom `um_metadata` table path) escape hooks to match. Both already exist.

## Testing

Coverage is manual since the change is a single opt-in filter.

**Test 1: backward compatibility**
1. Load the members directory on master and on this branch.
2. Compare every select/multiselect filter dropdown.
3. Result: options byte-identical when no dev hooks the new filter.

**Test 2: computed field populates the dropdown**
1. Add to a must-use plugin:
   ```php
   add_filter( 'um_member_directory_filter_values_hospital_list', function ( $values, $attrs ) {
       return array( 'Hospital A', 'Hospital B', 'Hospital C' );
   }, 10, 2 );
   ```
2. Reload the directory.
3. Result: the `hospital_list` dropdown lists all three hospitals.

**Test 3: search side, standard path**
1. Also hook `um_query_args_hospital_list__filter` to return a `meta_query` clause.
2. Select "Hospital A" in the dropdown and submit.
3. Result: results filter per the developer's callback (or empty if no stored meta matches — dev decides).

**Test 4: empty array from filter** — `return array()` hides the dropdown, same as if the DB had no rows.

**Test 5: multiselect** — flat array `array( 'X', 'Y' )` renders as multi-selectable options.

**Test 6: meta-table path** — pair with `um_query_args_hospital_list__filter_meta` to drive the custom `um_metadata` table search.

**Programmatic:**
- `php -l` on the modified file: clean
- `vendor/bin/phpcs` on the modified file: 0 new errors/warnings (file has 286 errors / 154 warnings pre-existing on master, all unchanged)


ashubawork:
@nikitasinelnikov 
The extension point looks useful, but I would frame this as an enhancement rather than a full fix for #596.

This filter lets developers supply synthetic dropdown values when the field values are computed and not stored in `wp_usermeta`, which solves the “empty dropdown” part of the problem. However, selecting one of those values will still use the existing member directory query path unless the developer also hooks `um_query_args_{$field}__filter` or `um_query_args_{$field}__filter_meta`. So the PR does not make computed fields searchable by itself; it only provides the missing dropdown values hook.

Before merging, I would suggest:
- pass more context to the filter, e.g. `$filter`, `$directory_data`, `$default_value`, `$admin`;
- update the wording so it is clear this supplies filter options, while the search query still needs to be wired separately.

The core idea is good and low-risk when unused, but the PR should avoid implying that `um_profile_{$key}__filter` automatically works for directory search after this change.

PR #1848, opened 2026-07-22
fix(members): add per-role permission to approve/deny registrations

### LINKED ISSUE #598: Add an option to give users registrations approval/denial rights to others than admins
There's a real problem with the current feature of approval/denial of new user's registration. It can only be done by administrators.

When a website is large, new users/members are coming more and more frequently. So most of the time, admins **need** to delegate the approval/denial to other team member(s) or moderators.
Problem, by default Ultimate Member allows only "administrators" to use the new users approval/denial features. And this is a real issue as:
- real admin(s) may just not have the time to do that so delegating this task may be very important
- but forcing to upgrade members to become (full) admins just to do this moderation task, just to approve/deny accounts, is far too dangerous. They would suddenly have access to ALL the website administration, they could break everything, the general settings, the theme, the plugins, even kick-off the main admin, etc. :-1: 

So a solution needs to be found to give this specific right to other user(s).

Add the following option to Ultimate Member > User Roles > edit role > Administrative permissions: "`Can approve/deny newly registered members? Checkbox ticked to say Yes`"

This option would give access to:
- wp-admin (equivalent to the 1st checkbox in Administrative permissions), to be able to view the list of users to approve/deny
- edit other member accounts (equivalent to the 3rd checkbox in Administrative permissions), to be able to update users status
- the content of the "Info" pop-up on users' record to display all the specific information a user has fill-in to register
- the dedicated functionalities on member's profiles, "Approve Membership", "Reject Membership", "Put as Pending Review" & "Deactivate this account" to ease the work
- and the Email notifications when a new member has registered and the approval is pending to know there's something to do

**Describe alternatives you've considered**
Currently the only thing we can do is to tick the 1st and 3rd checkboxes of the "Administrative permissions" section of other roles, like the Editor. But this is not enough. It does not provide:
- the content of the "Info" pop-up on users' record to display all the specific information a user has fill-in to register
- the dedicated functionalities on member's profiles, "Approve Membership", "Reject Membership", "Put as Pending Review" & "Deactivate this account" to ease the work
- and the Email notifications when a new member has registered and the approval is pending to know there's something to do

**Additional context**

faisalahammad:
## Summary

Adds a per-role permission `Can approve/deny newly registered members?` so non-admin roles (e.g. moderators) can approve, reject, put on hold, reactivate, or deactivate member registrations, view the registration Info popup, and receive pending-approval email notifications, without being granted the WordPress `edit_users` capability.

Closes #598

### `includes/admin/class-admin.php`
**Before:** `_um_can_approve_members` not registered in `$this->role_meta`, so its value is not sanitized on save.
**After:** New entry `_um_can_approve_members => array( 'sanitize' => 'bool' )` in the map, so role saves are sanitized via the existing pipeline.

### `includes/class-config.php`
**Before:** Default role metadata had no entry for the new key (existing installs default to 0 once added).
**After:** Every default role gets `_um_can_approve_members => 0`. Administrator gets `1`.

### `includes/admin/templates/role/admin-permissions.php`
**Before:** Role admin had no checkbox for the new permission.
**After:** New checkbox `Can approve/deny newly registered members?` placed directly under the existing `Can edit other member accounts?` checkbox. Tooltip explains the access it grants.

**Why:** This is the user-facing surface for the new permission.

### `includes/common/class-users.php` (central gate)
**Before:** `can_current_user_edit_user()` only checked `edit_users` capability and role meta.
```php
if ( ! current_user_can( 'edit_users' ) && ! $role->has_cap( 'edit_users' ) ) {
    return false;
}
```
**After:** Also returns true when the new permission is set.
```php
if ( ! current_user_can( 'edit_users' )
    && ! $role->has_cap( 'edit_users' )
    && ! UM()->roles()->um_user_can( 'can_approve_members' )
) {
    return false;
}
```

**Why:** 13+ call sites route through this gate. One extension covers them all.

### `includes/admin/class-actions-listener.php`, `class-users-columns.php`
**Before:** Standalone `current_user_can( 'edit_users' )` gates rejected moderator users before the central gate could run.
**After:** Each standalone gate now also accepts `UM()->roles()->um_user_can( 'can_approve_members' )`.

### `includes/frontend/class-users.php` (profile + member directory action dropdown)
**Before:** `$can_edit_users` only honored `edit_users`.
**After:** Also true for the new permission.

### `includes/admin/core/class-admin-builder.php` (Info popup AJAX)
**Before:** `dynamic_modal_content()` required `manage_options`; the per-case gate for `um_admin_review_registration` used a dead `current_user_can('administrator')` check.
**After:** Top-level gate keeps `manage_options` requirement, but bypasses it specifically for the `um_admin_review_registration` action when the caller has `edit_users` or the new permission. Per-case gate simplified to `um_can_view_profile()` check (consistent with other cases).

**Why:** The Info modal needs to be reachable by moderators, but other AJAX modal actions (form builder, field editor) must stay admin-only.

### `includes/core/um-actions-register.php`
**Before:** Pending-registration notification (`notification_review`) only went to `um_multi_admin_email()`.
**After:** After admin emails, also enqueues the same notification to any user whose role has the new permission, via a new `um_get_moderator_emails()` helper.

### `includes/um-short-functions.php`
**Before:** No helper to discover moderator emails.
**After:** New `um_get_moderator_emails()` returns the union of user emails whose priority role has `_um_can_approve_members = 1`. Uses `$wpdb->esc_like()` for the SQL LIKE pattern. Filterable via `um_get_moderator_emails`.

**Why:** Needed by the notification recipient extension. Smallest possible implementation, no caching added (YAGNI until traffic proves it).

### `includes/core/um-actions-access.php`
**Before:** Moderator viewing a pending profile would be redirected home.
**After:** Moderator allowed to view pending profiles.

**Why:** Without this, the new permission would let moderators click `Approve` but never see the profile it lives on.

**Before:** Member directory restricted to `approved` users when the viewer is not admin.
**After:** Moderators see all users (including pending) so the dropdown actions work on the full set.

### `includes/admin/class-site-health.php`
**Before:** Debug dump listed known role-meta keys.
**After:** New key added so admins can see its value in Site Health.

## Testing

**Test 1: Moderator can approve/deny a pending registration**
1. Create a new role `Moderator`. Tick `Can access wp-admin?` and `Can approve/deny newly registered members?`. Save.
2. Create a user, assign to `Moderator`, log in.
3. Register a new member through a UM registration form set to `Status: Pending Review`.
4. As moderator: open wp-admin Users, bulk-select the new pending user, choose `Approve Membership` under Ultimate Member bulk actions. Apply.
5. Result: user moves to `Approved` status. Admin and moderator both see the new state.

**Test 2: Moderator can open the Info popup and use the profile action dropdown**
1. As moderator, hover a pending user row in wp-admin Users. Click `Info`.
2. Result: registration data modal opens.
3. As moderator, open the same pending user's profile (frontend). Click the profile action dropdown.
4. Result: `Approve Membership`, `Reject Membership`, `Put as Pending Review`, `Reactivate this account`, `Deactivate this account` are all present and each works.

**Test 3: Notification recipient**
1. As moderator logged in, register a new member through a `Pending Review` form.
2. Result: admin receives `New account requires review` email. Moderator user `test-mod` also receives the same email.

**Test 4: Backward compat**
1. As Administrator, perform the full test 1+2 flow.
2. Result: behavior identical to pre-fix. No regressions.
3. Create a role with the new checkbox off. Assign a test user. Confirm no approval buttons, no Info modal, no notification email.

## Files Touched

- `includes/admin/class-actions-listener.php`
- `includes/admin/class-admin.php`
- `includes/admin/class-site-health.php`
- `includes/admin/class-users-columns.php`
- `includes/admin/core/class-admin-builder.php`
- `includes/admin/templates/role/admin-permissions.php`
- `includes/class-config.php`
- `includes/common/class-users.php`
- `includes/core/class-member-directory.php`
- `includes/core/um-actions-access.php`
- `includes/core/um-actions-register.php`
- `includes/frontend/class-users.php`
- `includes/um-short-functions.php`


ashubawork:
@nikitasinelnikov 
This can probably be implemented outside of core as an extension-level moderation workflow using existing hooks for displaying custom actions and custom action handlers for approve/reject.

I would avoid adding `_um_can_approve_members` directly to core in the current form, because the existing core permission helper `can_current_user_edit_user()` is too broad and is reused by deactivate/reactivate/resend activation/put pending flows.

PR #1849, opened 2026-07-22
fix(forms): add button class filter for 3 forms

### LINKED ISSUE #627: Option to change login and register button attributes e.g. class or id
Hello,

can you please give me as a developer to option to change the buttons on the register and login page? I really need this because I've created my own button class and I don't want to override every single button.

Thanks! 
A hard working developer

faisalahammad:
## Summary

Add `apply_filters` hook for primary submit button class on profile, password change, and account forms. Register, login, and password reset already had this filter (added in 2.10.5). Three forms were missing it.

Fixes #627

### includes/core/um-actions-profile.php
**Before:** `class="um-button"` hardcoded in template.
**After:** `um_profile_form_primary_btn_classes` filter, default `array( 'um-button' )`.
**Why:** Match existing convention, same filter signature as login/register.

### templates/password-change.php
**Before:** `class="um-button"` hardcoded in both branches (Set password / Change password).
**After:** `um_password_change_form_primary_btn_classes` filter.
**Why:** Password change form had no hook to modify button class.

### includes/core/class-account.php
**Before:** `class="um-button"` hardcoded.
**After:** `um_account_form_primary_btn_classes` filter with `` parameter so devs can target specific account tabs.
**Why:** Account form had no hook. Extra `` param enables tab-specific targeting.

## Testing

**Test 1: Default output unchanged**
1. Load register, login, profile edit, password reset, password change, and account pages.
2. Inspect the submit button HTML.
Result: `class="um-button"` on all forms (same as before).

**Test 2: Profile filter works**
1. Add filter: `add_filter( 'um_profile_form_primary_btn_classes', fn($c) => array_merge($c, ['my-class']) );`
2. Load profile edit page, inspect submit button.
Result: `class="um-button my-class"`.

**Test 3: Account filter with tab targeting**
1. Add filter checking `$tab_id === 'password'`.
2. Load account Password tab, inspect submit button.
Result: extra class appears only on Password tab, not on other tabs.


ashubawork:
@nikitasinelnikov I think that this is a useful commit

PR #1851, opened 2026-07-22
fix(access): allow content restriction on login/register pages

### LINKED ISSUE #636: Unable to restrict access to login/register for logged in users
**Expected behavior**
Redirect to configured URL in "UM Content Restriction" (Pages) when logged in.

- [x] I can reproduce this bug consistently using the steps below.

**How to reproduce?**
1. Login as Admin (WP)
2. Go to "Pages" -> "Login" (UM Login) -> "Edit"
3. Scroll down to "UM Content Restriction"
4. Enable checkbox "Restrict access to this content?"
5. Set "Who can access this content?" to "Logged out Users"
6. Set "What happens when users without access tries to view the content?" to "Redirect user"
7. Set "Where should users be redirected to?" to "Custom URL"
8. Set "Redirect URL" to _any public URL of your page_
9. Logout
10. Navigate to UM-Login page
11. Copy URL from browser addressbar (F6, ctrl+c)
10. Login (UM) - **as non admin**
11. Paste URL to browser addressbar (F6, ctrl+v)
12. Hit return
13. See your configured redirect **not** happening

**Why is this happening?**
Because there are explicit exceptions to check for access of UM-Core sites:

- https://github.com/ultimatemember/ultimatemember/blob/master/includes/core/class-access.php#L391
- https://github.com/ultimatemember/ultimatemember/blob/master/includes/core/class-access.php#L547

**Possible fixes**
- Remove exceptions for login/register (tested - working)
  - The option "Login page" in "UM Content Restriction" -> "Where should users be redirected to?" has to be removed while restricting access to "UM Login"
  - Potentially problematic because there are several default redirects to "UM-Login" (has to be tested)
- Remove "UM Content Restriction" section in "Pages" -> "Edit" for UM-Core pages (please dont!)

faisalahammad:
## Summary

Content restriction rules set on Login and Register pages were silently ignored for logged-in users. Two hardcoded exclusion gates in the access class skipped those core pages before reading any restriction meta, making the restriction settings no-ops.

## Changes

**get_post_privacy_settings()** — Remove `login` and `register` from the core-page exclusion list. The access check can now read the restriction metadata on those pages.

**template_redirect()** — Remove `login` and `register` from the early-return skip. The redirect logic now evaluates their restriction rules.

Excluded pages (unchanged): account, logout, password-reset, and logged-in user profile pages.

## Testing

1. Set Login page restriction to "Logged out Users" + redirect to custom URL.
2. Log in as non-admin → visit Login page URL → redirects to custom URL.
3. Logged-out user visits same URL → still sees login form.
4. Repeat on Register page with same config.
5. Verify account, logout, and password-reset pages still work without restriction interference.

## Notes

Restriction is opt-in per page. Nothing changes unless the admin explicitly enables it. Redirects go to user-defined custom URLs, not core pages — no circular redirect risk.

Fixes #636


ashubawork:
@faisalahammad 
Thanks for the PR. I agree with the underlying problem: restriction settings configured directly on the Login/Register pages should not be silently ignored.

However, we cannot accept this implementation because it removes the Login/Register exclusions from `template_redirect()` too broadly. That path is not limited to per-page restriction settings; it also runs the old/global access restriction logic. With global restricted access enabled and no custom access redirect URL configured, Ultimate Member falls back to the Login page as the redirect target. If the Login page itself is no longer excluded, this can result in a redirect-to-self loop.

So the fix needs to be narrower. We should allow individual restriction metadata on Login/Register pages to be evaluated, but keep those pages protected from the global restriction redirect flow, or add an explicit self-redirect guard.

For now, we are going to close this PR and handle the issue with a safer, more targeted change.

PR #1853, opened 2026-07-22
fix(templates): unify theme override path with templates/ subdir

### LINKED ISSUE #641: Location of members.php in Child Theme Directories
With reference to your documentation:
https://docs.ultimatemember.com/article/1516-templates-map

The only way I am able to make UM pick up that modified version is if I place it here:
wp-content/themes/yourtheme-child/ultimate-member**/templates**/members.php

And not the documented location:
wp-content/themes/yourtheme-child/ultimate-member/members.php

faisalahammad:
## Summary

Ultimate Member had two different theme override paths that disagreed on whether the `templates/` subdirectory is required. Sub-templates (members-header, members-grid, members-list, members-pagination, searchform, login-to-view, profile/posts, profile/comments) only resolved through the flat `ultimate-member/.php` path while the main template required `ultimate-member/templates/`. This PR unifies all four template loaders on the canonical path and keeps the legacy flat path as a backward-compatible fallback. It also accepts parent-theme overrides and provides a controlled allow-list for trusted custom template directories.

Fixes #641

### `includes/class-functions.php` — `UM()->locate_template()` + `UM()->get_template()`

**Before**
```php
$template = locate_template( array(
    trailingslashit( 'ultimate-member' . DIRECTORY_SEPARATOR . $path ) . $template_name
) );
```

**After**
```php
$candidates = array();
if ( ! $path ) {
    $candidates[] = trailingslashit( 'ultimate-member' . DIRECTORY_SEPARATOR . 'templates' ) . $template_name;
    $candidates[] = trailingslashit( 'ultimate-member' . DIRECTORY_SEPARATOR ) . $template_name;
} else {
    $candidates[] = trailingslashit( 'ultimate-member' . DIRECTORY_SEPARATOR . $path . DIRECTORY_SEPARATOR . 'templates' ) . $template_name;
    $candidates[] = trailingslashit( 'ultimate-member' . DIRECTORY_SEPARATOR . $path . DIRECTORY_SEPARATOR ) . $template_name;
}
$template = locate_template( $candidates );
```

The loader now accepts canonical and legacy overrides in both the active child theme and parent theme. The include guard uses the filterable `um_allowed_template_directories` list and rejects paths outside trusted directories. When `$path` is derived from a `$basename` argument, the matching extension directory `WP_PLUGIN_DIR/<path>/templates/` is allowed automatically, so templates shipped by add-ons keep working without registering through the filter.

### `includes/core/class-shortcodes.php` — `Shortcodes::load_template()`

The loader probes child and parent canonical paths first, then child and parent legacy flat paths, and uses the same allow-list guard before inclusion.

### `includes/common/class-theme.php` — `Theme::get_custom_templates_list()`

The scanner probes child-theme and parent-theme canonical paths first, then child-theme and parent-theme legacy flat paths, uniformly for every file, so Site Health detects the same override layouts as the loader.

### `changelog.txt`

Added a `= 2.13.1 =` section with a Bugfixes entry.

## Testing

1. Copy `templates/members.php` to `child-theme/ultimate-member/templates/members.php`, add a marker, visit members directory page, and confirm marker shows.
2. Copy `templates/members-list.php` to `child-theme/ultimate-member/templates/members-list.php`, change the empty-state string, visit a directory page showing no users, and confirm the new string appears.
3. Move override to flat `child-theme/ultimate-member/members-list.php`, confirm it still overrides (BC).
4. Place both canonical and legacy copies with different markers, confirm canonical wins.
5. Place an override in the parent theme, confirm it loads when the child has no matching copy.
6. Visit Tools -> Site Health, confirm UM templates panel detects the override.
7. Attach a filter rewriting `um_get_template` to `/etc/passwd`, confirm no include happens and a development notice is generated.
8. Add a trusted directory through `um_allowed_template_directories`, confirm a template from it can be loaded.
9. Place a random `anything.php` in the override folder, confirm no warnings or silent include.
10. Load a template through an extension `$basename` (for example an add-on calling `UM()->get_template( 'file.php', 'ultimate-member-pro/templates' )` or any add-on shipping `wp-content/plugins/<addon>/templates/`), confirm it loads without registering `um_allowed_template_directories`.
11. Place an override in the parent theme only, visit Tools -> Site Health, confirm the UM templates panel reports it (parent-theme override detection matches the loader).

## Backward compatibility

- Existing overrides at `ultimate-member/templates/<file>.php` continue to work.
- Existing overrides at the legacy flat `ultimate-member/<file>.php` continue to work as a fallback.
- Parent-theme canonical and legacy overrides are supported through the same lookup and guard.
- Extension templates under `WP_PLUGIN_DIR/<basename>/templates/` are allowed automatically; no add-on change is required.
- Integrations with custom template paths can add trusted directories with `um_allowed_template_directories`.
- No public API surface changed. No new actions or globals were added.

## Compatibility

- WordPress 6.2+ (current `Requires at least`)
- PHP 7.0+ (current `Requires PHP`)


ashubawork:
@faisalahammad 

Thanks for the PR. The direction makes sense: template override lookup should be consistent, and keeping the legacy flat path as a fallback is useful for backward compatibility.

I do not think this should be merged as-is yet. The new allow-list guard only permits the plugin templates directory and `get_stylesheet_directory()` override paths. However, `locate_template()` can also resolve files from the parent theme via `get_template_directory()`. With this patch, a valid parent-theme override could be found and then rejected by the include guard, which would be a regression.

There is also a compatibility concern around the `um_get_template` / `um_locate_template` filters. They have historically allowed integrations to provide template files from custom locations. Blocking every path outside the UM plugin templates and theme override folders may be a valid security hardening decision, but it is still a behavior change. If we keep this guard, we probably need either a documented allow-list filter for trusted template directories or a clearer migration path for add-ons.

One more thing: for non-empty `$path`, the fallback to `ultimate-member/<template_name>` looks a bit too broad and could cause filename collisions between template namespaces. I would prefer keeping fallback lookup scoped to the current path/namespace unless there is a specific compatibility case for the global flat path.

faisalahammad:
Addressed the review points.

- Added parent-theme canonical and legacy template directories to the include allow-list. `locate_template()` results from `get_template_directory()` are now accepted.
- Added the `um_allowed_template_directories` filter for trusted custom directories. Rejected paths are skipped with a `_doing_it_wrong()` notice.
- Reused the allow-list in `Shortcodes::load_template()` and added parent-theme probing there.
- Removed the global `ultimate-member/<file>` fallback when `$path` is non-empty. Path-scoped legacy lookup remains.
- Merged current master and moved the changelog entry to 2.13.1.

Checks pass for PHP syntax, changed-line PHPCS, and CodeRabbit review. POT regeneration was not available because this checkout has no WordPress installation or i18n WP-CLI command.

faisalahammad (commented):
Implemented the requested template path fixes.

- Parent-theme canonical and legacy override directories are now accepted.
- `um_allowed_template_directories` provides a controlled extension point for trusted custom paths.
- Both template loaders use the shared realpath allow-list check.
- Non-empty template namespaces no longer fall back to the global flat path.
- Current master was merged and the changelog was updated.

PHP syntax checks, changed-line PHPCS checks, independent verification, and CodeRabbit review passed.

ashubawork:
@faisalahammad 
Thanks, this version is much closer. The canonical `ultimate-member/templates/` lookup with the legacy flat path as a fallback looks good, and the include allow-list is a good hardening step for filtered template paths.

I still see two blockers before merge.

First, the allow-list does not appear to account for plugin/extension template directories resolved through the `$basename` argument. `UM()->get_template()` derives `$path` from `$basename`, and `locate_template()` can fall back to `WP_PLUGIN_DIR . '/' . $path . '/templates/' . $template_name`. With the new `is_allowed_template_path()` guard, those templates may be located but then rejected unless the extension manually adds its directory via `um_allowed_template_directories`. That would be a backward compatibility break for legitimate extension templates.

Second, the PR description says the Site Health/template scanner supports parent-theme overrides, but the `Theme::get_custom_templates_list()` change still checks only `get_stylesheet_directory()`. That means the loader and scanner can disagree: a parent-theme override may load, but Site Health will not report it.

faisalahammad:
@ashubawork

Quick note. I am replacing my previous long comment, please read this one instead.

Both blockers from your 2026-09-07 review are fixed at d47fb378d (head now f26a3ee37 after the translation bot push).

1. Allow-list now accepts extension templates. `get_allowed_template_directories()` and `is_allowed_template_path()` take an optional `$path`. When non-empty, `WP_PLUGIN_DIR/<path>/templates/` is appended before the `um_allowed_template_directories` filter, and `get_template()` forwards `$path`. Add-ons passing their own `$basename` keep working. Other custom directories (mu-plugins, symlinked plugin roots) still go through the filter.

2. `Theme::get_custom_templates_list()` now probes parent canonical and parent legacy paths in the same order as `Shortcodes::load_template()`. Site Health matches the loader.

Also fixed two stray-tab indents on the guard blocks from the previous commit. PHPCS is back to pre-PR baseline, no new violations. CodeRabbit reports 0 findings. Head SHA: f26a3ee37.

Could you take another look when you have a moment? Requested formal reviewers do not work for the PR author on this org, so I am pinging you directly here.

ashubawork:
@faisalahammad 
Thanks, the latest revision addresses my previous blockers. Passing $path into is_allowed_template_path() preserves extension templates under WP_PLUGIN_DIR/<path>/templates/, and the Site Health scanner now checks child/parent theme overrides for both canonical and legacy paths, so loading and scanning are aligned.

I think this is acceptable after small cleanup:
- Please do not translate identifier strings like `um_get_template` and `load_template`; pass them as literals to `sprintf()` so they do not become separate msgids.
- Optional hardening: validate/sanitize `$path` before adding `WP_PLUGIN_DIR/<path>/templates` to the allowed directories, so dot segments cannot broaden the allowed path.

Once cleaned up, this looks mergeable to me.

faisalahammad:
Done, thank you for the review.

- The filter and function names are no longer wrapped in `__()`, so they do not add translatable strings.
- `get_allowed_template_directories()` now skips a `$path` that contains `..` or a null byte.
- Latest master is merged. The only conflict was in generated translation files.

@ashubawork please re-review when you have time.

PR #1855, opened 2026-07-24
fix(profile): allow multiple profile forms with role-specific option

### LINKED ISSUE #838: Allow multiple profile forms to display with the role-specific option
When you create two Profile forms with the same role-specific settings, it only displays 1 profile form. I suggest adding a filter hook to allow us to show the forms in this kind of setup.

faisalahammad:
## Summary

Add the `um_allow_multi_role_profile` filter so plugins can opt in to rendering more than one profile form with role-specific settings on the same page. The filter wraps the suppression check that has dropped the second profile form shortcode since the 2020 dedup commit. Default behavior stays unchanged.

Fixes #838

### `includes/core/class-shortcodes.php`

**Before:**
```php
} elseif ( $this->profile_role !== $args['role'] ) {
    ob_get_clean();
    return '';
}
```

**After:**
```php
} elseif ( $this->profile_role !== $args['role'] ) {
    /**
     * Filters the ability to render a role-specific profile form
     * after another role-specific profile form has already rendered
     * on the same page.
     *
     * Default behavior keeps the original shortcode suppression so
     * only the first matching profile form renders. Return `true`
     * to opt in to rendering the second form on the page.
     *
     * The filter only fires when an earlier role-specific profile
     * form on the same page has already set `$this->profile_role`.
     *
     * @since 2.12.2
     * @hook um_allow_multi_role_profile
     *
     * @param {bool}  $allow Render profile form or not. Default false (suppress).
     * @param {array} $args  Shortcode arguments.
     *
     * @example <caption>Allow multiple profile forms with role-specific settings on the same page.</caption>
     * add_filter( 'um_allow_multi_role_profile', '__return_true' );
     */
    $allow = apply_filters( 'um_allow_multi_role_profile', false, $args );
    if ( ! $allow ) {
        ob_get_clean();
        return '';
    }
    $this->profile_role = $args['role'];
}
```

**Why:** the issue reporter asked for a filter hook to make the role-suppression opt-in. Wrapping the existing check keeps ship behavior unchanged and exposes it for theme/plugin developers. Updating `$this->profile_role` after the filter opts in keeps suppression logic consistent for any number of subsequent forms on the same page, regardless of order.

### Version bumps
- `ultimate-member.php`: Version 2.12.1 → 2.12.2
- `readme.txt`: Stable tag 2.12.1 → 2.12.2
- `changelog.txt`: new 2.12.2 section noting the filter

## Testing

Setup: create two Profile forms (Form A and Form B), each with "Apply custom settings to this form" enabled and `role-specific` set to different roles (e.g. Customer and Author). Place `[ultimatemember form_id="A"]` and `[ultimatemember form_id="B"]` on the same page.

**Test 1: default behavior unchanged**

1. Visit the page as a user with the Customer role.
2. Do not add the filter.

Result: only Form A renders. Form B's shortcode output is empty. Matches the original 2020 dedup behavior.

**Test 2: opt-in via filter**

1. Add to a theme or custom plugin: `add_filter( 'um_allow_multi_role_profile', '__return_true' );`
2. Reload the page as the same user.

Result: both forms render.

**Test 3: PHP lint**

`php -l includes/core/class-shortcodes.php` reports no syntax errors.

## Backend only

No UI changes, no JS/CSS touched. No screenshots needed.

## Compatibility

- Backward compatible: filter default is `false`. Existing sites unchanged.
- PHP 7.0+ compatible, no new PHP syntax.
- `@since 2.12.2` aligned with `Version:` and `Stable tag` bumps in the same PR.


ashubawork:
@nikitasinelnikov 
If we keep this PR, I suggest adjusting the implementation so the new filter only controls whether an additional role-specific profile form may render, not whether role validation is skipped.

I updated the logic so the current profile user’s roles are always checked before `um_allow_multi_role_profile` is applied. This prevents `__return_true` from accidentally rendering a role-specific form for a role the profile user does not have.

I also normalized the form role setting to a sorted array before comparing it with the previously rendered role-specific form. This makes comparisons stable when the role setting is stored as either a string or an array, or when the same role set appears in a different order.

The filter now receives extra context: `$form_roles` and `$current_user_roles`, so custom implementations can make a more precise decision.

PR #1856, opened 2026-07-24
fix(directory): use translatable defaults for empty Members Directory results text

### LINKED ISSUE #1181: Members Directory number of Profile hits should be translatable
**Expected behavior**
Members Directory number of Profile hits should be translatable if several languages are used at the site.

**Describe the bug**
UM is always using the user text entered in the the textboxes at UM Member Directories "Results Text", "Single Result Text" and "Custom text if no users were found" even if the textboxes are empty.

**Expected behavior**
Use the translated message if the textbox is empty.

**UM Support Forum Issue**
https://wordpress.org/support/topic/issues-with-members-page-translations/

**Solution**
Install this code snippet by adding it to your active theme’s functions.php file
or use the “Code Snippets” Plugin

https://wordpress.org/plugins/code-snippets/

Customize the template `members-header.php`

**Code snippet**
```
add_filter( 'um_ajax_get_members_response', 'um_ajax_get_members_response_headers', 100, 2 );

function um_ajax_get_members_response_headers( $member_directory_response, $directory_data ) {

    if ( empty( $directory_data['header'] )) {
        $string = __( '{total_users} Members', 'ultimate-member' );
        $member_directory_response['pagination']['header'] = UM()->Member_Directory()->convert_tags( $string, $member_directory_response['pagination'] );
    }

    if ( empty( $directory_data['header_single'] )) {
        $string = __( '{total_users} Member', 'ultimate-member' );
        $member_directory_response['pagination']['header_single'] = UM()->Member_Directory()->convert_tags( $string, $member_directory_response['pagination'] );
    }

    if ( empty( $directory_data['no_users'] )) {
        $member_directory_response['pagination']['no_users'] = __( 'We are sorry. We cannot find any users who match your search criteria.', 'ultimate-member' );
    }

    return $member_directory_response;
}
```

**Template**
```
<?php if ( ! defined( 'ABSPATH' ) ) exit; ?>


<script type="text/template" id="tmpl-um-members-header">
	<div class="um-members-intro">
		<div class="um-members-total">
			<# if ( data.pagination.total_users == 1 ) { #>
				{{{data.pagination.header_single}}}
			<# } else if ( data.pagination.total_users > 1 ) { #>
				{{{data.pagination.header}}}
			<# } else if ( data.pagination.total_users == 0 ) { #>
				{{{data.pagination.no_users}}}
			<# }#>
		</div>
	</div>

	<div class="um-clear"></div>
</script>
```

faisalahammad:
## Summary
When the admin leaves the **Results Text** or **Single Result Text** textboxes blank on a Members Directory, the frontend renders an empty string instead of the existing translatable defaults. This change falls back to the same translated strings already shipped by the plugin (`{total_users} Members` / `{total_users} Member`), so Polylang/WPML/TranslatePress translators can localize them on translate.wordpress.org.

Fixes #1181

**Before** (in `calculate_pagination()` and `predefined_no_caps()`):
```php
$pagination_data['header']        = $this->convert_tags( $directory_data['header'], $pagination_data );
$pagination_data['header_single'] = $this->convert_tags( $directory_data['header_single'], $pagination_data );
```

**After**:
```php
$header        = ! empty( $directory_data['header'] )        ? $directory_data['header']        : __( '{total_users} Members', 'ultimate-member' );
$header_single = ! empty( $directory_data['header_single'] ) ? $directory_data['header_single'] : __( '{total_users} Member', 'ultimate-member' );
$header        = apply_filters( 'um_member_directory_default_header', $header, $directory_data, $pagination_data );
$header_single = apply_filters( 'um_member_directory_default_header_single', $header_single, $directory_data, $pagination_data );

$pagination_data['header']        = $this->convert_tags( $header, $pagination_data );
$pagination_data['header_single'] = $this->convert_tags( $header_single, $pagination_data );
```

**Why:** Mirrors the existing fallback pattern at `templates/members.php:180` which already does this for the `_um_directory_no_users` field. The default strings are reused from `UM()->config()->core_directory_meta['members']` — they were already registered as translatable at plugin install; the only behavior gap was that empty admin input overrode them.

## New filters

- `um_member_directory_default_header` — receives `$header`, `$directory_data`, `$pagination_data`. Theme/plugin authors can restore the legacy "blank = blank" behavior with `add_filter( 'um_member_directory_default_header', '__return_empty_string' )`.
- `um_member_directory_default_header_single` — same shape.

## Testing

**Test 1: empty textbox falls back to translatable default**
1. WP Admin → Members Directory.
2. Clear the **Results Text** and **Single Result Text** textboxes, save.
3. Visit the Members page on the frontend.
4. Expected: the directory header reads `50 Members` (or whatever count). When the count is 1, the header reads `1 Member`.

**Test 2: custom admin text not translated (BC preserved)**
1. Enter `Found profiles!` into **Results Text**, save.
2. Visit the page.
3. Expected: shows `Found profiles!` — admin input is not run through `__()`, no XSS regression.

**Test 3: translators can localize the fallback**
1. Clear the textboxes (Test 1 setup).
2. Use Polylang/WPML/TranslatePress to translate the `ultimate-member` text domain.
3. Switch language on the front-end.
4. Expected: the default `{total_users} Members` is replaced with the language-specific translation.

## BC notes

- Admin-entered non-empty text untouched — never wrapped in `__()`/`sprintf()`.
- Whitespace-only input still bypasses the fallback (matches the existing `no_users` pattern at `templates/members.php:180`).
- Theme/plugin authors who relied on "blank = blank" can opt out via the new filter.

## Notes for reviewer

The two patched sites are the only PHP writers of `pagination_data['header']` / `header_single`. All AJAX and initial-render JSON flows route through `calculate_pagination()` and the capability-checked early-return path `predefined_no_caps()`, so both paths are covered.


ashubawork:
@faisalahammad 

Thanks for the PR. The multilingual use case is valid, but we should not change the existing behavior of an intentionally empty Results Text / Single Result Text field.

At the moment, leaving the field blank is documented in the admin UI as a way to hide the result text (“Leave this blank to not show result text”). With this patch, an intentionally blank saved value would start rendering the default `{total_users} Members` / `{total_users} Member` text, which would be a behavior change for existing sites.

I think the fix needs to preserve the current semantics:

- an intentionally empty saved field should still hide the result text;
- a directory that has no saved custom value should be able to use the translated default;
- custom non-empty admin text should remain unchanged.

So we need a BC-safe way to distinguish “no saved value, use default” from “saved empty value, hide text”.

faisalahammad:
Hi @ashubawork, thanks for catching this, you are right.

Fixed: changed the empty check from `!empty()` to `array_key_exists()` in `calculate_pagination()` and `predefined_no_caps()` (`includes/core/class-member-directory.php`). Now:

- If Results Text / Single Result Text was ever saved (even blank), the directory keeps the exact current behavior: blank stays hidden. This covers your case, an admin who cleared the field on purpose sees no change.
- Only a directory whose meta key was never saved at all falls back to the translated default.

I also found the real reason the built-in "Members" directory could never be retranslated: its default text got resolved once by `__()` at plugin activation and saved as plain text in postmeta, so language switchers had nothing to hook into. Updated `includes/core/class-setup.php` so that directory no longer bakes the string in, it stays a live `__()` call. This closes the original translatability report (#1181) for the default directory without touching any already-saved directory's data.

Pushed to this branch. Branch is also rebased on latest master now (it was ~6 weeks behind, which is why it showed as not mergeable).

PR #1857, opened 2026-07-25
fix(directory): support tabindex on select and multiselect fields

### LINKED ISSUE #1333: UM resetting SELECT statements to TABINDEX= -1
**Expected behavior**
User code adding TABINDEX to INPUT, TEXTAREA and SELECT statements should not be changed by UM code.

**Describe the bug**
UM is changing after the user code is setting a positive TABINDEX value to a SELECT statement the value to -1
https://imgur.com/a/nnykwA3

**To Reproduce**
Try this plugin “Tabindex Settings”
https://github.com/MissVeronica/um-tabindex-settings

**Expected behavior**
No SELECT statement changes to TABINDEX= -1

**Support Forum issue**
https://wordpress.org/support/topic/how-to-redesign-registration-page/

faisalahammad:
## Summary
Support custom tabindex on UM select/multiselect fields via the field data array, so plugins can set tab order (e.g. MissVeronica/um-tabindex-settings). Also fix a namespace resolution bug in the bundled action-scheduler wrapper that prevented plugin activation when the library files are present.

Fixes #1333

### includes/core/class-fields.php
**Before:** select and multiselect tags rendered without any tabindex attribute from field data. Any tabindex set via DOM manipulation was overwritten by the bundled Select2 at init time.

**After:** reads `$data["tabindex"]` from the field data array, validates with `is_numeric()`, sanitizes with `absint()` and `esc_attr()`, and renders it as a tabindex attribute on the underlying `<select>` element. Select2's existing init mechanism transfers the original value to the visible `<span.select2-selection>` widget.

**Why:** Select2 caches the original tabindex from the source element at init time and moves it to the visible widget. Setting tabindex after init is too late. Rendering it from field data before Select2 runs lets Select2's built-in mechanism preserve the custom value.

### includes/action-scheduler/class-init.php
**Before:** `as_next_scheduled_action()`, `as_enqueue_async_action()`, and 6 other `as_*()` functions were called without a leading backslash from the `um\action_scheduler` namespace.

**After:** all 8 calls use `\as_*()` to resolve to the global namespace.

**Why:** PHP resolves unqualified function calls to the current namespace. The `as_*()` functions are defined in the global namespace by the bundled Action Scheduler library. Without the backslash prefix, PHP looks for `um\action_scheduler\as_next_scheduled_action()`, which does not exist, causing a fatal error on plugin activation.

## Testing

**Test 1: Tabindex on select field**
1. Add a filter in theme `functions.php`: `add_filter("um_get_field_select", fn($d) => array_merge($d, ["tabindex" => 5]));`
2. Create a registration form with a select field (e.g. Country).
3. View the form on the front end. Inspect the dropdown with browser DevTools.
4. The hidden `<select>` should have `tabindex="-1"` (Select2 default), the visible `<span.select2-selection>` should have `tabindex="5"`. Tab through the form and verify the select is at position 5.

**Test 2: Backwards compatibility (no tabindex set)**
1. Remove the filter from step 1. Reload the form.
2. Neither the `<select>` nor the Select2 widget should render a tabindex attribute. No PHP warnings in debug.log.

**Test 3: Plugin activation**
1. Install the patched version with Composer dependencies on a fresh test site.
2. Activate the plugin. Verify there's no fatal error about `as_next_scheduled_action()`.

**Test 4: Multiselect support**
1. Add a multiselect field to a form.
2. Add filter: `add_filter("um_get_field_multiselect", fn($d) => array_merge($d, ["tabindex" => 10]));`
3. Same behavior as Test 1 — tabindex 10 should appear on the visible widget.

## Screenshots

| Before | After |
|--------|-------|
| SCR-20260725-lanb | SCR-20260725-lbpd |


ashubawork:
@nikitasinelnikov 
The tabindex fix is valid, but I adjusted the implementation slightly before keeping it.

Instead of building the tabindex attribute inline in both select and multiselect render branches, I moved it into a small `get_tabindex_attr()` helper. This keeps the markup code cleaner and avoids duplicating the same validation/escaping logic.

I also changed the value handling from `absint()` to `(int)`. HTML `tabindex` supports negative values and `0`, so `absint( -1 )` would incorrectly render `tabindex="1"`. With `(int)`, values like `-1`, `0`, and positive custom tab indexes are preserved correctly.

The Action Scheduler namespace fix is already applied locally and looks valid as a separate change, but it is unrelated to the tabindex issue and should ideally stay as a separate commit/PR.

PR #1858, opened 2026-07-25
fix(form-builder): hide _editable on Register forms

### LINKED ISSUE #1297: Improve field settings visibility in Form Builder
I recommend that we hide field settings such as 'Can user edit this field` which doesn't work with Register forms since 1.0.0

I saw that we are moving to blocks for building forms. Temporarily, we can disable these unused options with available filter hook `um_core_fields_hook` when displaying the field settings.

I'll create a PR for this

faisalahammad:
## Summary
Hides the "Can user edit this field?" control in the form builder's field modal when the parent form's mode is `register`. The setting is meaningless on registration (user has no field to edit yet) and was always visible, leaving admins confused about a toggle that does nothing.

This follows the same approach already used in this file at `case '_validate':` to hide the validation control on Login forms for the built-in credential fields.

Fixes #1297

### `includes/admin/core/class-admin-metabox.php`
**Before:**
```php
case '_editable':
    // Make a new field editable by default.
    if ( false === $this->in_edit ) {
        $this->edit_mode_value = true;
    }
```

**After:**
```php
case '_editable':
    // Field cannot be edited during registration; control is meaningless on Register forms.
    if ( array_key_exists( 'mode', $form_data ) && 'register' === $form_data['mode'] ) {
        break;
    }

    // Make a new field editable by default.
    if ( false === $this->in_edit ) {
        $this->edit_mode_value = true;
    }
```

**Why:** The form mode is already loaded into `$form_data['mode']` at the top of `field_input()`, so the guard adds no new query and matches the same idiom used for the `_validate` setting on Login forms. Profile and Login modes continue to render the control normally; only Register is suppressed.

## Testing

**Test 1: Register form — control must be hidden**
1. Go to Ultimate Member > Forms. Edit a form whose Mode is Register.
2. Click "Add Field" or edit an existing field.
3. Open the field modal and scroll through the columns.
Result: the "Can user edit this field?" checkbox is NOT present.

**Test 2: Profile form — control must still appear**
1. Edit a form whose Mode is Profile.
2. Open a field's modal.
Result: the "Can user edit this field?" checkbox is visible, existing values round-trip correctly.

**Test 3: Login form — control must still appear**
1. Edit a form whose Mode is Login.
2. Open a field's modal.
Result: the "Can user edit this field?" checkbox is visible.

**Test 4: Existing field settings preserved**
1. On a Profile form, edit a field, note its `_editable` value, save.
2. Reopen the same field.
Result: saved value matches what was shown before save.

## Screenshot
image


ashubawork:
@nikitasinelnikov 
I agree with the direction of this PR: `_editable` is a profile-edit setting, so showing “Can user edit this field?” on Register forms is misleading.

I adjusted the condition slightly so the control is rendered only for Profile forms, instead of hiding it only for Register forms. The setting is consumed by profile edit logic (`um_can_edit_field()` and field rendering checks when the form mode is `profile`), so it is also not meaningful for Login forms.

Updated logic:
if ( empty( $form_data['mode'] ) || 'profile' !== $form_data['mode'] ) {
    break;
}

This keeps the fix aligned with where `_editable` actually affects behavior, while preserving the existing Profile form behavior.

PR #1859, opened 2026-07-26
fix(validation): use WP_oEmbed for youtube_video URL validation

### LINKED ISSUE #1331: Validate URLs via WP_oEmbed::get_provider($url)
Avoid using 3rd-party regular expressions for validation media links and use WordPress native WP_oEmbed class.

Ref: https://github.com/ultimatemember/ultimatemember/pull/1326#pullrequestreview-1664925590

faisalahammad:
## Summary

Replace the hand-rolled `UM()->validation()->is_url()` + `um_youtube_id_from_url()` regex in the `youtube_video` field validator with WordPress's native `WP_oEmbed::get_provider()`. This lets WordPress maintain the provider URL regex list instead of us keeping a separate pattern in sync.

Fixes #1331

### includes/core/um-actions-form.php

**Before:**
```php
case 'youtube_video':
    if ( ! UM()->validation()->is_url( $submitted_data[ $key ] ) || false === um_youtube_id_from_url( $submitted_data[ $key ] ) ) {
        UM()->form()->add_error( $key, sprintf( __( 'Please enter a valid %s URL', 'ultimate-member' ), $array['label'] ) );
    }
    break;
```

**After:**
```php
case 'youtube_video':
    if ( '' !== $submitted_data[ $key ] ) {
        $wp_oembed = _wp_oembed_get_object();
        $provider  = is_string( $submitted_data[ $key ] ) ? $wp_oembed->get_provider( $submitted_data[ $key ], array( 'discover' => false ) ) : false;
        if ( ! $provider || false === strpos( $provider, 'youtube.com' ) ) {
            UM()->form()->add_error( $key, sprintf( __( 'Please enter a valid %s URL', 'ultimate-member' ), $array['label'] ) );
        }
    }
    break;
```

**Why:** The original PR #1326 review suggested using `WP_oEmbed::get_provider()` as a future improvement. WordPress already maintains regex patterns for YouTube, Vimeo, SoundCloud, and other oEmbed providers, so we can reuse that logic. Network discovery is disabled to avoid outgoing requests during form validation. The provider URL is also checked against `youtube.com` so non-YouTube oEmbed providers do not accidentally pass validation for this field type.

## Testing

**Test 1: Valid YouTube watch URL**
1. Submit `https://www.youtube.com/watch?v=dQw4w9WgXcQ`
Result: validation passes, value saved.

**Test 2: Valid Shorts URL**
1. Submit `https://www.youtube.com/shorts/dQw4w9WgXcQ`
Result: validation passes, value saved.

**Test 3: Invalid URL**
1. Submit `https://example.com/foo`
Result: form error displayed.

**Test 4: Empty field**
1. Leave the field blank
Result: validation passes (optional field).

**Test 5: Non-YouTube oEmbed URL**
1. Submit `https://vimeo.com/12345`
Result: validation fails (provider URL does not contain `youtube.com`).


ashubawork:
@faisalahammad 
Thanks for the PR. The direction makes sense: using WordPress’ oEmbed provider list is better than maintaining our own media URL regex.

However, I do not think we should merge this as-is because it only updates validation. The profile rendering path still uses `um_youtube_id_from_url()` to extract the video ID before building the embed URL. This can make validation and rendering accept different URL formats. For example, a URL accepted by `WP_oEmbed::get_provider()` may be saved successfully but then fail to render if `um_youtube_id_from_url()` cannot extract the ID from that format.

I think the fix should keep validation and rendering aligned. Either the extractor should be updated to support the same YouTube URL formats that validation accepts, or both validation and rendering should share one helper that validates the provider and returns the embed-ready video ID.

The provider check should be stricter than `strpos( $provider, 'youtube.com' )`; parsing the provider URL host would be safer and clearer.

faisalahammad:
Thanks for the feedback. Both points are addressed in the latest push.

Validation and rendering now share a single helper `um_youtube_video_id_from_url()` in `includes/um-short-functions.php`. The helper:
1. Asks `_wp_oembed_get_object()->get_provider( $url, array( 'discover' => false ) )` so no network call is made.
2. Parses the returned provider URL host with `wp_parse_url( $provider, PHP_URL_HOST )`, strips a leading `www.`, and requires an exact match against `youtube.com` (lowercased).
3. Extracts the video ID via the existing `um_youtube_id_from_url()` pattern, which now also matches the `/live/<id>` path.

`um_submit_form_errors_hook_` (validation) and `um_profile_field_filter_hook__youtube_video()` (rendering) both call this helper, so accepted URLs always render and rejected URLs never reach the renderer. The loose `strpos` check is gone.

The existing `um_youtube_id_from_url()` function is kept public and unchanged in behavior aside from the `/live/` addition, for any third-party callers.

ashubawork:
@faisalahammad Thanks, this version is much better. I like that validation and rendering now share the same `um_youtube_video_id_from_url()` helper, and the provider check is safer now that it parses the provider host and compares it to `youtube.com` exactly. Disabling oEmbed discovery is also the right choice for form validation.

@nikitasinelnikov The only remaining behavior question is `youtube.com/embed/<id>` URLs. The old `um_youtube_id_from_url()` supported `/embed/`, but WordPress `WP_oEmbed::get_provider()` does not include the embed URL pattern, so this PR now rejects URLs that UM previously accepted. If we want to preserve backward compatibility, we should add an explicit fallback for YouTube embed URLs after strict host validation, or document this as an intentional behavior change.

PR #1860, opened 2026-07-28
fix(uploader): clean stale photo thumbnails

### LINKED ISSUE #1027: Profile Photo Thumbnail Size differences
**Describe the bug**
Uploading a profile photo not possible to resize to all thumbnail sizes will not be displayed after replacing a profile photo with all thumbnail sizes available.

- [ ] I have deactivated other plugins and confirmed this bug occurs when only Ultimate Member plugin is active.
- [ ] This bug happens with a default WordPress theme active, or [UM Theme](https://ultimatemember.com/theme/).
- [X] I can reproduce this bug consistently using the steps above.

**To Reproduce**
1. Set "Profile Photo Thumbnail Sizes" to: 40, 80, 190, 250 px
2. Upload a profile photo with size like 1024 x 1024 px
3. The large profile photo is displayed in the profile and the member directory
4. Upload a new profile photo for replacement of the large photo with size like 190 x 190 px
5. Show the profile, still with the first profile photo 

**Expected behavior**
The smaller profile photo to replace the larger profile photo.

**Possible bypass solution**
User must remove the larger profile photo first then a second step uploading the smaller profile photo.

**Investigation**
When uploading the smaller profile photo only thumbnail sizes possible to create are replaced and the 190x190 and 250x250 sizes with the larger photo are not being removed in the upload process.
Removing the `profile_photo-250x250.jpg` will display the smaller profile_photo and the `profile_photo-190x190.jpg` is still the larger photo and should also have been replaced as the upload was 190x190 px.

**Possible UM solution**
Removing all profile photo files before upload in `function ajax_image_upload()`

```
				if ( ! is_array( $_FILES[ $id ]['name'] ) ) {
					UM()->files()->delete_core_user_photo( $user_id, 'profile_photo' );
					UM()->uploader()->replace_upload_dir = true;
					$uploaded = UM()->uploader()->upload_image( $_FILES[ $id ], $user_id, $id );
					UM()->uploader()->replace_upload_dir = false;
					if ( isset( $uploaded['error'] ) ) {
						$ret['error'] = $uploaded['error'];
					} else {
						$ret[] = $uploaded['handle_upload'];
					}

				}
```
I have no solution for the 190x190 case. Probably an equal sign missing.

faisalahammad:
## Summary

When a smaller photo replaces a larger one, `WP_Image_Editor::multi_resize()` cannot generate thumbnails larger than the new source. Old, larger thumbnails stay on disk and keep being served by `um_get_avatar_uri()`, so the previous photo still shows after replacement.

Fix removes leftover `profile_photo-*` / `cover_photo-*` files in the user upload directory with `wp_delete_file()` before `multi_resize()` regenerates the new set. Same cleanup applied to both profile photo and cover photo for consistency.

Fixes #1027

### `includes/core/class-uploader.php`

**Before:** `profile_photo()` and `cover_photo()` called `multi_resize()` which only generated thumbnails bounded by the new source dimensions. Stale thumbnails from prior uploads persisted.

**After:** Before each `multi_resize()` call, glob `wp-content/uploads/{user_id}/{key}-*` and delete each match with `wp_delete_file()`. Glob pattern uses `{key}-*` (with dash) so the canonical `profile_photo.jpg` / `cover_photo.jpg` are not touched; only thumbnail files like `profile_photo-190x190.jpg` match.

```php
// inside profile_photo() and cover_photo(), before multi_resize()
$dir = UM()->files()->upload_basedir . $user_id . DIRECTORY_SEPARATOR;
if ( is_dir( $dir ) ) {
	$stale = glob( $dir . $key . '-*' );
	if ( is_array( $stale ) ) {
		foreach ( $stale as $old ) {
			if ( is_file( $old ) ) {
				wp_delete_file( $old );
			}
		}
	}
}
```

**Why:** `multi_resize()` is a one-way generator bounded by source dimensions. The only safe way to remove leftovers is to delete them explicitly before regenerating. Doing it in `profile_photo()` / `cover_photo()` (not `ajax_image_upload()`) avoids clobbering the temp file used by the cropper staging flow.

## Testing

Manual repro steps in `TESTING_INSTRUCTIONS.md` at the repo root of the fix branch. Highlights:

1. Settings > General > Uploads, set Profile Photo Thumbnail Sizes = `40,80,190,250`.
2. Upload a 1024x1024 profile photo, verify all 4 thumbnails exist in `wp-content/uploads/{user_id}/`.
3. Replace with a 190x190 photo.
4. Verify no stale `-250x250.jpg` remains. After fix, only sizes the new source can produce stay on disk.

Same scenario for cover photo.

Verified: PHPCS clean on changed lines (WordPress-Extra standard). No automated test framework exists in this repo.

PHP tested on the version currently in master. No new runtime dependencies.


ashubawork:
@faisalahammad 
The fix addresses a real issue: when a smaller profile/cover photo replaces a larger one, `multi_resize()` does not remove previously generated larger thumbnails, and UM may continue serving those stale files.

I think the approach is generally correct, but I would not merge it exactly as-is. Please move the duplicated cleanup block into a small helper.

Since this code deletes files via `glob()`, I would also add an explicit guard in the helper so it only runs for `profile_photo` and `cover_photo`, even though the normal call path already passes those keys.

faisalahammad (commented):
@ashubawork Updated as requested.

- Moved the duplicated cleanup block into a private helper `delete_stale_thumbnails()` in `includes/core/class-uploader.php`.
- The helper has an explicit guard now. It only runs for `profile_photo` and `cover_photo`.
- Both call sites in `profile_photo()` and `cover_photo()` call the helper before `multi_resize()`.

Fixed in commit f4db3e7a0. PHPCS is clean on the changed lines. Please take another look.

PR #1863, opened 2026-08-09
fix(assets): inline Select2 i18n locale to prevent AMD reordering crash

### LINKED ISSUE #1755: Select2 i18n Fix
**UM Support Forum issue**
https://wordpress.org/support/topic/uncaught-typeerror-at-en-js/

**Solution**
https://github.com/rytisp/um-select2-i18n-fix

faisalahammad:
## Summary
The Select2 i18n locale file (`assets/libs/select2/i18n/{locale}.js`) was enqueued as a separate `<script src>` tag with `select2` as a dependency. Script reordering optimizers (Autoptimize, WP Rocket, LiteSpeed, TranslatePress, and others) can move the locale tag ahead of `select2.full.min.js`. When that happens, the minified `en.js` IIFE's `e.define(...)` call runs before `jQuery.fn.select2.amd` is defined, throwing `Uncaught TypeError: Cannot read properties of undefined (reading 'define')`.

UM now inlines the locale code on the `select2` handle via `wp_add_inline_script()`. The inline payload is printed in the same `<script>` block as `select2.full.min.js`, after it, with no reorderable boundary.

Fixes #1755

### `includes/common/class-enqueue.php` (`register_select2()`)
**Before:**
```php
if ( file_exists( UM_PATH . 'assets/libs/select2/i18n/' . $locale . '.js' ) ) {
    wp_register_script( 'um_select2_locale', $libs_url . 'select2/i18n/' . $locale . '.js', array( 'jquery', 'select2' ), '4.0.13', true );
    self::$select2_handle = 'um_select2_locale';
}
```
**After:**
```php
if ( file_exists( UM_PATH . 'assets/libs/select2/i18n/' . $locale . '.js' ) ) {
    static $select2_locale_inline_added = false;

    $locale_path = UM_PATH . 'assets/libs/select2/i18n/' . $locale . '.js';
    $locale_src  = file_get_contents( $locale_path );
    if ( false !== $locale_src && '' !== trim( $locale_src ) && ! $select2_locale_inline_added ) {
        $inline                      = 'if ( jQuery && jQuery.fn && jQuery.fn.select2 && jQuery.fn.select2.amd ) { (function(){ ' . $locale_src . ' })(); }';
        $select2_locale_inline_added = (bool) wp_add_inline_script( 'select2', $inline, 'after' );
    }
}
```
**Why:**
- `wp_add_inline_script()` attaches the locale code to the existing `select2` print group, so it is emitted in the same `<script>` block as `select2.full.min.js`. No second request, no reorderable boundary.
- The wrapper guard (`if ( jQuery && ... .amd ) { ... }`) skips the `e.define(...)` call entirely if the full Select2 build ever fails to load, which the bundled minified `en.js` does not do on its own (its minified guard is broken — the `if` only protects the `var e` assignment, not the subsequent `e.define(...)` call).
- The `static $select2_locale_inline_added` flag prevents duplicate append when `register_select2()` runs more than once in a request (e.g. when `frontend/class-enqueue.php` deregisters and re-registers the script).
- `self::$select2_handle` stays at its default `'select2'`, so all four downstream consumers (`um_scripts` in `includes/frontend/class-enqueue.php` and `includes/admin/class-enqueue.php`, plus `um_admin_forms` and `um_admin_builder`) keep resolving to the full Select2 build, which now also carries the inline locale payload.

## Testing
**Test 1: Original error is gone after the fix**
1. On `master` (before the fix), activate Autoptimize with JS aggregation, load a page that uses Select2 (registration form, edit profile, member directory filter), and open DevTools Console. Reload a few times until the optimizer places `i18n/en.js` ahead of `select2.full.min.js`. The console reports `Uncaught TypeError: Cannot read properties of undefined (reading 'define')` at the `en.js` URL.
2. Switch to this branch (or install `ultimate-member-fix-1755.zip`), reload. The error no longer appears.
3. View source on a Select2 page. There is no separate `<script src=".../i18n/en.js">` tag. The locale code is printed inline inside the `<script>` block for `select2.full.min.js`, immediately after it.

**Test 2: Localization still works**
1. Settings -> General -> Site Language -> switch to a locale with a bundled Select2 i18n file (e.g. `de_DE`, `fr_FR`, `es_ES`, `it_IT`, `pt_BR`, `ja`).
2. Open a registration form or edit profile page that uses a Select2 field. Confirm the dropdown placeholder and no-results message render in the selected language.

**Test 3: Regression check**
1. Open a page that uses Select2 (registration form, edit profile, member directory filter). Confirm dropdowns open, search works, placeholders render.
2. WP Admin -> Form Builder. Select2-backed fields still work.
3. With Autoptimize still active, inspect the network panel. The locale inline script always appears immediately after the `select2.full.min.js` `<script>` tag, regardless of optimizer behavior.

## Notes
- The public filter `um_dequeue_select2_scripts` behavior is unchanged.
- The WooCommerce coexistence branch (dequeue + re-register) is unchanged.
- The `um_select2_locale` handle is no longer registered. It was never part of UM's public API and no code outside `register_select2()` references it (verified by grep across `includes/`). If an extension ever enqueued its own script on that handle to override the locale, it would silently stop working post-fix; none of the bundled 63 locale files contain a `</script>` sequence, so inline emission is safe.


ashubawork:
@nikitasinelnikov 
I agree with the direction of this PR: inlining the Select2 locale after the `select2` handle is a good way to avoid script reordering issues caused by optimization plugins.

I adjusted the duplicate-prevention logic. The original `static $select2_locale_inline_added` flag can become stale because `register_select2()` may deregister and re-register the `select2` handle in the WooCommerce / `um_dequeue_select2_scripts` branch. In that case, inline data added to the old handle is dropped, but the static flag would still prevent adding the locale to the newly registered handle.

Instead, the updated version checks the current `select2` handle’s `after` inline data for a small locale marker before adding the inline script. This avoids duplicates during normal repeated calls, while still allowing the locale to be reattached after a deregister/register cycle.

I also changed the JS guard to use `window.jQuery` so the guard itself cannot throw a `ReferenceError` if the global `jQuery` variable is unavailable.

ashubawork:
@faisalahammad 
Thanks for the PR. The direction looks good, but we cannot push a follow-up commit to your branch because maintainer edits are disabled.

Could you either enable “Allow edits by maintainers” for this PR, or update the duplicate-prevention logic so it checks the current `select2` handle inline data instead of using a static flag?

The static flag can become stale if `register_select2()` deregisters and re-registers the `select2` handle. In that case, inline data added to the old handle is dropped, but the static flag still prevents adding the locale to the new handle.

faisalahammad:
@ashubawork Both points addressed in a1f5f2f: (1) dropped the static `` flag — `register_select2()` now inspects the current `select2` handle's `extra['after']` inline data for a `um-select2-locale:<locale>` marker, so the locale reattaches after the WooCommerce / `um_dequeue_select2_scripts` deregister-register cycle while duplicates stay prevented on repeated calls; (2) the JS guard now uses `window.jQuery` (with `fn`/`select2`/`amd` checks), so the guard itself cannot throw a `ReferenceError` when the global identifier is unavailable. Ready for re-review.

faisalahammad:
@ashubawork Force-pushed merge of `origin/master` into PR branch: ccca43fd6. Resolved conflicts in `languages/ultimate-member-en_US.po` and `languages/ultimate-member.pot` by taking the regenerated master versions; the Select2 locale fix in `includes/common/class-enqueue.php` is preserved unchanged. Please re-review.

PR #1866, opened 2026-08-09
fix: email template data URI CSS stripping on save (#1430)

### LINKED ISSUE #1430: Support more CSS attributes & values in Email Templates
I propose to support the `backgound:url()` and others  in the `style=""`  attribute. 

Right now, when you add this to the email template > Welcome Email: https://gist.github.com/champsupertramp/32e6e60ce4c20f7878ccc05b88a24b2d

The preview shows the image, but when you save it and come back to the editor, the `background: url()`  is removed.

faisalahammad:
## Summary

An admin who adds `background: url('data:image/jpeg;base64,...')` to an email template inline `style=""` (for example in the Welcome Email) loses that `url()` declaration after Save. The preview shows it (it uses the raw editor value), but the saved template file has it stripped.

This is caused by WordPress core, not by Ultimate Member's tag filtering. `wp_kses()` passes every `style` value to `safecss_filter_attr()`, which rebuilds its own allowed-protocol list from `wp_allowed_protocols()`. That function uses a static cache that only applies the `kses_allowed_protocols` filter before `wp_loaded`, so at admin save time the `data` scheme the plugin merges into its own `$protocols` argument never reaches the URL check. The whole `url(data:...)` declaration is then dropped.

Fixes #1430

### `includes/admin/core/class-admin-settings.php`

`Admin_Settings::save_email_templates()` now runs template content through a data-aware sanitizer before writing the theme template file.

**Before:**
```php
$content  = wp_kses( stripslashes( $settings[ $template ] ), 'post', $protocols );
```

**After:**
```php
$content  = $this->sanitize_email_template( stripslashes( $settings[ $template ] ), $protocols );
```

The new `sanitize_email_template()` protects each inline `style=""` value with a placeholder token, runs the content through `wp_kses( ..., 'post' )` for tag and attribute level checks, then restores each value through `sanitize_email_template_style()`.

`sanitize_email_template_style()`:
- Allows `url()` schemes from the core defaults plus `data`, `blob` and `cid`.
- Decodes CSS escapes (`\6a avascript`) and HTML entities before scheme and `expression()` checks, so obfuscated `javascript:` and `expression()` are still stripped.
- Keeps relative and protocol-relative URLs, which carry no scheme.
- Splits declarations on top-level semicolons only, so `;base64` payloads inside a `data:` URI are not cut.
- Object attribute breakout is blocked by `esc_attr()` on restore.

**Why:** core `safecss_filter_attr()` cannot see the `data` scheme (see root cause above), so it must be scrubbed with the data-aware list. The tag/attribute allowlist is unchanged.

## Testing

**Repro (the bug):**
1. Ultimate Member > Settings > Email, open Welcome Email.
2. Paste an ``.
3. Save, then reopen the editor.

Result: `background: url('data:...')` persists in the saved template. Before the fix it was stripped.

**Regression:**
- Relative, parent-relative, protocol-relative, `http(s)` and `data` URIs (quoted and unquoted) all kept.
- SVG data URI with `;base64` payload and following declarations kept intact.
- Plain CSS and custom properties (`--brand`) kept.
- `<script>`, `onclick`, `url(javascript:)`, CSS `expression()`, and their CSS-escaped variants all still stripped.

**Environment:** WP 6.2+, PHP 7.0 (repo minimum). No UI change, so no screenshots.

## Notes

- Only `includes/admin/core/class-admin-settings.php` is modified. No new hooks and no global protocol or `safe_style_css` filters were added.
- Templates that were stripped by this bug before the fix must be edited and saved once more to restore the `url()` declarations. No migration.


ashubawork:
@faisalahammad 

Thanks for working on this. The reported issue is valid, but I do not think this implementation is safe enough to merge as-is.

The current patch hides inline `style` attributes from `wp_kses()` and then restores them after running a custom CSS sanitizer. That means we bypass WordPress core’s `safecss_filter_attr()` behavior for email templates. The custom sanitizer checks URL protocols and `expression()`, but it does not preserve WordPress’ allowed CSS property list, so properties that core would normally strip could be saved.

I am also concerned that the patch allows broad `data:`, `blob:`, and `cid:` URLs. The issue only needs support for image data URIs in CSS backgrounds, so the exception should be much narrower, e.g. only `data:image/png`, `data:image/jpeg`, `data:image/gif`, and maybe `data:image/webp` with base64 payloads. I would avoid allowing `data:image/svg+xml`, generic `data:`, and `blob:` here.

A safer approach would be to keep WordPress’ normal `wp_kses()` / safe CSS filtering and only make a very targeted exception for safe image data URIs, for example by temporarily replacing allowed `url(data:image/...;base64,...)` values with placeholders before `wp_kses()` and restoring only those exact values afterward. That way we solve the email background-image issue without replacing core CSS sanitization.

faisalahammad:
Updated the sanitizer to keep WordPress core KSES/safecss_filter_attr in the save path. The workaround now replaces only narrowly validated base64 data:image/png, jpeg, gif, or webp URLs inside CSS url() values with per-save inert HTTPS tokens, then restores only tokens that survive inside style attributes. Generic data, SVG, blob, cid, javascript, malformed URLs, and disallowed CSS properties remain handled by core filtering. PHP syntax and focused sanitizer checks pass; CodeRabbit reports no findings.

PR #1868, opened 2026-08-10
Add admin notice for outdated overridden templates

### LINKED ISSUE #1662: Include "template is out of date" check as an UM update package.

Admins are not re-checking their templates after an UM version update neither at UM Settings -> Advanced -> Override templates nor doing a  Site Health report.

None of these steps are mentioned in the very old [UM Documentation](https://docs.ultimatemember.com/article/201-how-to-update-your-site)  (December 15,  2021) and there are many issues posted in the support Forum because of outdated templates.

Add the "template is out of date" check during UM plugin update and report any issues immediately for admin action.

faisalahammad:
## Summary

Adds a persistent admin notice that warns when any overridden Ultimate Member template is out of date, with a direct link to Settings > Advanced > Override Templates. The notice recomputes on each admin load and disappears automatically once the admin updates (or removes) the stale override.

Closes #1662

### `includes/admin/core/class-admin-notices.php`

**Before:**
Outdated template detection only surfaced inside Site Health and Settings > Advanced > Override Templates, both easy to miss after a plugin update.

**After:**
```php
/**
 * Checking if there are outdated overridden templates in the active theme
 */
public function outdated_templates_notice() {
	// create_list() runs on every admin_init, including admin-ajax.php requests from non-admins.
	// render_notices() shows this notice to the users with the 'manage_options' capability only, so skip
	// the template scan below for everybody else.
	if ( ! current_user_can( 'manage_options' ) ) {
		return;
	}

	$has_outdated = false;
	foreach ( UM()->common()->theme()->build_templates_data() as $template ) {
		if ( isset( $template['status_code'] ) && 0 === (int) $template['status_code'] ) {
			$has_outdated = true;
			break;
		}
	}

	if ( ! $has_outdated ) {
		return;
	}

	// The first iteration of this notice was dismissible with the same key. A dismissal left by that
	// version would hide this non-dismissible notice, so drop it while outdated templates exist.
	$hidden_notices = get_option( 'um_hidden_admin_notices', array() );
	if ( is_array( $hidden_notices ) && in_array( 'outdated_templates', $hidden_notices, true ) ) {
		update_option( 'um_hidden_admin_notices', array_values( array_diff( $hidden_notices, array( 'outdated_templates' ) ) ) );
	}

	$override_url = admin_url( 'admin.php?page=um_options&tab=advanced&section=override_templates' );

	$allowed_html = array(
		'a' => array(
			'href' => array(),
		),
	);

	$this->add_notice(
		'outdated_templates',
		array(
			'class'       => 'notice-warning',
			// translators: %s: Override templates settings link.
			'message'     => '<p>' . wp_kses( sprintf( __( 'Your custom Ultimate Member templates are out of date and may affect functionality. Please <a href="%s">update your overridden templates</a>.', 'ultimate-member' ), esc_url( $override_url ) ), $allowed_html ) . '</p>',
			'dismissible' => false,
		),
		10
	);
}
```

Registered in `create_list()` next to the other condition-based notices (`exif_extension_notice`, etc.).

**Why:** The plugin already has the detection logic, but the result is only visible in Site Health and the Override Templates settings screen. Admins rarely re-check those spots after an update, so stale overrides go unnoticed until something breaks. A dashboard notice keeps the warning visible until resolved.

**Non-dismissible by design:** after the review discussion the notice is now non-dismissible. It can be hidden only by fixing the cause: once every theme override matches its core template version, the scan finds no outdated template and the notice is not registered anymore. A leftover dismissal from the first (dismissible) iteration of this PR is removed automatically while outdated templates exist.

## Testing

**Test 1: Outdated template triggers the notice**
1. Copy a core template (e.g. `templates/profile.php`) to `[theme]/ultimate-member/templates/profile.php`.
2. Remove or lower the `@version` header in the copied file so it is older than the bundled template.
3. Log in as an admin and load any wp-admin page.

Result: a `notice-warning` appears with a link to Settings > Advanced > Override Templates.

**Test 2: Notice auto-resolves when fixed**
1. From Test 1, raise the override `@version` header to match (or exceed) the bundled template.
2. Reload any wp-admin page.

Result: the notice disappears without any manual cleanup.

**Test 3: Notice is not dismissible**
1. From Test 1, load any wp-admin page and look at the outdated templates notice.
2. Reload the page several times.

Result: the notice has no dismiss button and keeps re-appearing on every admin load while the outdated override exists. It only disappears after the override is updated or removed (Test 2).

**Test 4: Leftover dismissal from the first iteration does not hide the notice**
1. Set the option so `outdated_templates` is in `um_hidden_admin_notices` (e.g. WP-CLI: `wp option update um_hidden_admin_notices '["outdated_templates"]' --format=json`).
2. With an outdated override in place, load any wp-admin page.

Result: the stale dismissal entry is removed and the notice is shown.

Verified: `php -l` clean, PHPCS clean on the changed lines, no new unit test coverage added (no existing tests cover `Admin_Notices`).


ashubawork:
@faisalahammad 
The feature direction is good and the issue is valid: outdated overridden templates are easy to miss if the warning only appears in Site Health or the Override Templates settings screen.

I do not think we should merge this as-is because the notice is dismissible with a static key (`outdated_templates`). UM stores dismissed notices in `um_hidden_admin_notices`, so once an admin dismisses this warning it will stay hidden even after a future UM update or after templates become outdated again. That weakens the main purpose of this PR.

Could you either make this notice non-dismissible, or make the dismissal context-aware, for example by including the UM version or a hash of the outdated template list in the notice key?

A couple of smaller notes:
- please escape the settings URL with `esc_url()` before output;
- consider softer wording such as “may affect functionality” instead of “may break functionality”;

faisalahammad:
@ashubawork Thanks for the review. Pushed b64aaa5.

**Dismissal is context aware now.** The key is `outdated_templates_` plus an md5 of the outdated template paths and their versions, so a dismissal hides only that exact situation. When an update makes templates stale again the key changes and the notice returns. Old keys get dropped so `um_hidden_admin_notices` does not grow.

Trade off: if a release changes no already outdated override, the key stays the same and the notice stays hidden. Narrower than a UM version key, which would re-nag every release. Say the word and I will switch.

Settings URL uses `esc_url()`. Text says "may affect functionality".

Two fixes in the same method while testing:

- Replaced the uncached `is_outdated_template_exist()` with `build_templates_data()`. `create_list()` runs on `admin_init`, so every wp-admin request, including logged out `admin-ajax.php` calls, did a recursive `scandir()` plus two file reads per override.
- Added a `manage_options` check. `render_notices()` gates on it but `create_list()` does not.

Also merged master to clear the translation file conflict.

ashubawork:
@faisalahammad 
Thanks, this update addresses the main concern from the previous review. The dismissal key is now context-aware, so the notice can reappear when the outdated template set or versions change after a plugin update. The capability guard, escaped URL, and softer wording also look good.

I only have a small cleanup note before merge:
In `cleanup_outdated_templates_dismissals()`, please match the generated notice prefix as `outdated_templates_` instead of `outdated_templates`. That keeps the cleanup scoped only to these fingerprinted notice keys and avoids removing any unrelated future notice that may share the broader prefix.

After that, I think this is good to merge.

faisalahammad:
Updated `cleanup_outdated_templates_dismissals()` to match `outdated_templates_` only. This keeps cleanup limited to fingerprinted outdated-template notice keys and avoids removing unrelated keys with the broader prefix. The fix is in commit `92227bb37`; manual testing is confirmed.

faisalahammad (commented):
The requested cleanup prefix fix is complete in `92227bb37`. The code now matches only `outdated_templates_` fingerprint keys. PHP syntax passed, CodeRabbit reported no findings, and manual testing is confirmed.

nikitasinelnikov:
Hi guys,

I'm not sure about to merge this PR or not. Previously we had the similar wp-admin notice but based on the option (bool - has outdated templates or not). See ref here https://github.com/ultimatemember/ultimatemember/blob/b40edd26e6b00370f0e85e1ce851df3756da384f/includes/admin/core/class-admin-notices.php#L779-L804

As soon as Site Health is stable we moved this functionality to it instead of the wp-admin notice because a lot active plugins makes the wp-admin top page unuseful. See my screen, it's 17'' laptop and 60% of the screen height are filled by the admin notices 
image

The notice was not dismissible, and it's ok. Outdated templates can be the major issue so I think this type of the wp-admin notice if we return it back should be still not dismissible and can be hidden only after all templates in theme have the same version as the templates in the UM Core.

I want to know your thoughts,
Best Regards!

faisalahammad:
@nikitasinelnikov Thanks for the context, the history helps a lot. I agree on the notice fatigue. My thoughts:

This notice is different from the marketing notices in your screenshot. It is a single notice that appears only for admins, and only on sites that actually have an outdated override. On a clean site it never renders. The scan is cached, so the cost per admin load is small.

I agree it should not be dismissible. The notice is non-dismissible now and follows your rule: it can be hidden only by fixing the cause. While any override is older than its core template the notice renders on every admin load. Once every override matches the core version the notice is not registered anymore.

The fingerprint key and the dismissal cleanup from the earlier rounds are removed, so the code is simpler now. I also drop a leftover dismissal saved by the first dismissible iteration of this PR, so that state cannot hide the notice.

Happy to keep it in Site Health only if you still prefer that, but I think the notice is worth it because Site Health is easy to miss after a plugin update. The updated commit is untested-by-you so far; a build zip is ready if you want to try it.

faisalahammad:
@nikitasinelnikov Updated and pushed in `7faab0ac2`.

The notice is now non-dismissible and stays visible until all overridden templates match the core versions. I also clear the old dismissal key from the first iteration so it cannot hide the warning. Manual testing is confirmed, CodeRabbit found no issues, and the PR is clean.

Please review again when you have time.

PR #1869, opened 2026-08-10
fix(profile): display multi-value fields as lists

### LINKED ISSUE #1353: Improve Profile page readability in view mode with unordered bulleted lists
Comma separated lists are difficult to read when each user option for `multiselect` and `checkbox` fields  are long text strings.

Replace current comma separated `implode` string for `multiselect` and `checkbox` fields with an unordered HTML bulleted list in the function `um_profile_field_filter_hook__`.

**Describe alternatives you've considered**
Code snippet for array formatting before the UM array `implode`.
```
add_filter( 'um_profile_field_filter_hook__', 'um_field_value_bullet_list', 10, 3 );

function um_field_value_bullet_list( $value, $data, $type ) {

    if ( $type == 'multiselect' || $type == 'checkbox' ) {

        $meta_value_list = '<ul>';
        foreach( $value as $meta_value ) {
            $meta_value_list .= '<li>' . $meta_value . '</li>';
        }
        $meta_value_list .= '</ul>';
        return $meta_value_list;
    }
    return $value;
}
```

**UM Support Forum issue**
https://wordpress.org/support/topic/checkbox-output-as-list/

faisalahammad:
## Summary
Add an opt-in setting to display Multi-select and Checkbox profile field values as an unordered bulleted list on the main profile page. This improves readability for profiles with several selected values while preserving current comma-separated output by default.

Fixes #1353

### Profile field display
**Before:**
```text
Value One, Value Two, Value Three
```

**After:**
```html
<ul class="um-field-list">
    <li>Value One</li>
    <li>Value Two</li>
    <li>Value Three</li>
</ul>
```

**Why:**
The new Appearance setting is disabled by default for backward compatibility. List items are escaped and passed through a restrictive KSES allowlist. Profile header/meta values and member directory output remain unchanged.

## Testing
**Test 1: Opt-in list display**
1. Install the built plugin package on WordPress 6.2+ with PHP 7.0+.
2. Create a profile form with Multi-select and Checkbox fields containing multiple long options.
3. Enable **Display values as a bulleted list on profiles** under **Ultimate Member > Settings > Appearance**.
4. View a profile with selected values.
Result: Each selected value displays as a separate unordered-list item.

**Test 2: Backward-compatible default**
1. Disable the new setting.
2. View the same profile.
Result: Values display as comma-separated text.

**Test 3: Scope and escaping**
1. Check another profile field type, profile header/meta values, and member directory values.
2. Use an option label containing HTML characters.
Result: Existing output remains unchanged outside the main profile field view, and HTML characters render as escaped text.

Automated checks passed: PHP syntax checks for all three changed files and `git diff --check`. CodeRabbit review reported zero findings. The repository has no usable PHPUnit test runner in this environment.


ashubawork:
@faisalahammad 
The feature direction is good. Displaying long checkbox/multiselect profile values as a list can improve readability, and keeping it opt-in with a default `false` setting preserves backward compatibility.

I would not merge this PR as-is yet.

Main concerns:
1. The PR includes unrelated `direct_db_value_cache` changes from another issue/PR. Please split those out.
2. The list rendering should not fall back to `explode( ', ', $res )`. Option labels can contain commas, so this can split values incorrectly. We should build the list only from the original array value when the submitted/stored field value is actually an array.
3. Empty values should be skipped after trimming.
4. It would be cleaner to move the `<ul>` rendering into a small helper instead of adding more formatting logic directly inside `view_field()`.

faisalahammad:
@ashubawork Thanks for the review. All four points are done in 08bc324.

1. Unrelated cache changes are split out. The `direct_db_value_cache` property, the `get_direct_db_value()` helper and both call sites are removed. `is_selected()` and `is_radio_checked()` are back to the original code.

2. The `explode( ", ", $res )` fallback is gone. The list is built from the stored array value now, so an option title with a comma inside stays in one list item.

3. Values that are empty after `trim()` are skipped. When every value is skipped the helper returns an empty string, so there is no empty `<ul></ul>` in the output.

4. The `<ul>` rendering moved to a new private helper `get_values_list( $key, $data, $type )` in `includes/core/class-fields.php`. `view_field()` only calls it and checks the result.

One extra note on point 2. While fixing it I found the array never reaches `view_field()`. In view mode `field_value()` returns early with `um_filtered_value()`, and `um_profile_field_filter_hook__` joins arrays with `", "` at priority 99. So a check on `is_array( $_field_value )` alone would make the setting dead code. The helper reads the stored array with `um_user( $key )` and then reuses the same steps the comma output uses to get the titles: `um_profile_field_filter_xss_validation()` for options validation and key to title mapping, `get_option_value_from_callback()` for callback based dropdowns, and `__()` for multiselect option translation. So both modes show the same titles and no string is split.

I also merged the latest `master` into the branch and fixed the conflict in the generated `.pot` and `.po` files by taking the `master` version. The PR is mergeable again.

Checks: `php -l` on the changed files, `phpcs` with `WordPress-Extra` shows no findings on the changed lines, and CodeRabbit reports no findings. Tested manually with the built package on a profile with multiselect and checkbox fields, including an option title with a comma and an option label with HTML characters.

ashubawork:
@faisalahammad 
Thanks, the updated implementation addresses my previous concerns: the unrelated cache change is gone, the list is no longer built by splitting the rendered comma-separated value, empty values are skipped, and the helper keeps the profile rendering logic much clearer.

I re-checked the sanitization flow as well. The generated list markup is not passed through the bio-specific `wp_kses()` block for checkbox/multiselect fields, and the individual list values are escaped with `esc_html()`, so this part looks fine to me.

After that cleanup, @nikitasinelnikov I think this PR is good to merge.

faisalahammad (commented):
Manual testing confirmed.

On the main profile page:
- With "Display values as a bulleted list on profiles" enabled, multi-select and checkbox values render as an unordered list.
- An option label containing a comma stays in one list item.
- An option label containing HTML characters is escaped.
- With the setting disabled, values return to comma-separated text.
- Profile header, profile meta, and member directory output are unchanged.

Headless checks pass: PHP syntax, git diff --check, PHPCS on the changed lines, and CodeRabbit. No regressions found.

Ready to merge.

PR #1870, opened 2026-08-10
fix(fields): cache duplicate usermeta checks

### LINKED ISSUE #1535: Redundant SQL calls when editing Profile page with callback dropdowns
**Expected behavior**
Redundant SQL calls should be minimized by redesign or caching procedures.

**Describe the bug**
Adding two dropdowns to select US state and US state county will give 49 plus 1849 SQL calls identical for the two meta_keys used for the callback dropdowns when editing the Profile page. Time calculation by Query Monitor to perform these SQL calls are 0.15 seconds

**To Reproduce**
1. Install Query Monitor
2. Implement "Dynamic Country-State-City Dependent Dropdown fields" from UM documentation 
3. Edit the Profile page
4. See Query Monitor Database Duplicate Queries

**Expected behavior**
Three possible solutions:

1. Save returned value from the two statements in `fields.php` for the `$key` value to be used for next calls.
`$direct_db_value = $wpdb->get_var( $wpdb->prepare( "SELECT ISNULL( meta_value ) FROM {$wpdb->usermeta} WHERE user_id = %d AND meta_key = %s", um_user( 'ID' ), $key ) );`

```
if ( ! isset( $this->is_selected_cache[$key] )) {
	$this->is_selected_cache[$key] = $wpdb->get_var( $wpdb->prepare( "SELECT ISNULL( meta_value ) FROM {$wpdb->usermeta} WHERE user_id = %d AND meta_key = %s", um_user( 'ID' ), $key ) );
}
$direct_db_value = $this->is_selected_cache[$key];
```

2. Change coding by using `um_user( $key )` instead of the "$wpdb->get_var( $wpdb->prepare( "SELECT ISNULL........." as the um_user  function is a very good cache.

3. Redesign the functions `is_selected` and `is_radio_checked` for consecutive calls

**Support Forum Issue**
Large callback dropdown question
https://wordpress.org/support/topic/using-an-api-with-choices-callback/

faisalahammad:
## Summary
Cache repeated usermeta existence checks while rendering profile fields. This removes duplicate SQL calls for every dropdown option while preserving default value behavior.

Fixes #1535

### `includes/core/class-fields.php`
**Before:**
```php
$direct_db_value = $wpdb->get_var( $wpdb->prepare( "SELECT ISNULL( meta_value ) FROM {$wpdb->usermeta} WHERE user_id = %d AND meta_key = %s", um_user( 'ID' ), $key ) );
```

**After:**
```php
$direct_db_value = $this->get_direct_db_value( $key );
```

The result is cached per user ID and meta key. Cached `NULL` values are retained with `array_key_exists()`, so fields without a usermeta row still show their configured default.

## Testing
**Test 1: Query Monitor duplicate queries**
1. Add state and county callback dropdown fields using the Dynamic Country-State-City setup.
2. Open the Ultimate Member Profile page in edit mode.
3. Inspect duplicate database queries in Query Monitor.
4. Confirm the `SELECT ISNULL( meta_value )` query runs once per meta key, not once per dropdown option.
5. Confirm saved values, empty saved values, and missing usermeta rows keep existing behavior.

Result: manual verification confirmed the duplicate queries were reduced as expected.

Automated checks passed:
- `php -l includes/core/class-fields.php`
- `git diff --check`
- CodeRabbit review passed with 0 findings.

No PHPUnit runner or Composer-installed PHPCS executable is available in this repository.


ashubawork:
@nikitasinelnikov 
This is a valid optimization for the issue #1535. The repeated `SELECT ISNULL(meta_value)` query is executed once per option during profile edit rendering, so caching it per user ID and meta key makes sense and should preserve the existing default value behavior.

A few small cleanup requests before merging:
1. Consider renaming `get_direct_db_value()`. The method does not return the actual meta value; it returns the direct DB existence/null check result used to decide whether the field default should apply.
2. The cache property can probably be `private` or `protected` instead of public, since it is an internal implementation detail.

nikitasinelnikov:
Hi @faisalahammad 
Could you change the base branch to `development/2.9.x` for your active PRs? After changing the base from the `master` branch to `development/2.9.x`, I cannot resolve the conflicts and merge the proper PRs where it's possible.


Btw, this PR is ok and I'll merge it manually
Best Regards!

PR #1872, opened 2026-08-17
fix(fields): migrate old country names on upgrade

### LINKED ISSUE #1722: Documentation: How to update a country name change
In the recent UM country name changes (Türkiye and North Macedonia) Users are left without any documentation on how to update their dropdowns in all UM Profile and Registration Forms and User saved home country names as UM development is not supplying any new release update job fixing these issues. 

These country name changes affects all UM sites with country name dropdowns regardless of which is the main site country. It's not an issue only for two specific countries.

faisalahammad:
## Summary
Adds a 2.12.2 upgrade package that renames the old country labels in saved data: form field options and defaults, conditional logic values, member directory default filters, and user metadata (including multiselect values and the submitted registration copy). The country renames shipped in 2.10.3 (Turkey to Türkiye) and 2.10.5 (Macedonia, the former Yugoslav Republic of to North Macedonia) changed the built-in list only, so forms and profiles kept the old values, and the profile display validation blanks saved values that are no longer in the options list (`um_profile_field_filter_xss_validation` in `includes/core/um-filters-fields.php`).
Fixes #1722

### New upgrade package `includes/admin/core/packages/2.12.2/`
`hooks.php`, `functions.php`, `init.php` following the existing 2.3.0/2.8.0 package pattern, run from the Ultimate Member > Upgrade admin page with paged AJAX (100 usermeta rows per request).

**Why:** select field values are stored as the option label string, and the form builder saves the options list into `_um_custom_fields` post meta at form save time, so persisted data keeps whatever names were current when it was saved.

Replacements applied, exact matches only:
```php
'Turkey'                                     => 'Türkiye',
'Macedonia, the former Yugoslav Republic of' => 'North Macedonia',
```

The routine updates:
- `_um_custom_fields` post meta on all forms (any status): `options`, `default`, conditional logic values, option keys in pair mode
- `um_fields` option: same walk for global custom fields
- `_um_search_filters` post meta on member directories with default filters limited by the old names
- usermeta rows for the affected metakeys plus the predefined `country` field and the `submitted` registration copy, updating the exact row via `update_metadata_by_mid()` and clearing the `um_cache_userdata_{id}` cache

Generic `custom_dropdown_options_source` callbacks are not used as a country-field detector. They can power any dynamic select/multiselect field, and callback-backed fields skip the strict static-option validation, so they are not added to the usermeta migration list unless a stored field config actually contains one of the old country values. If pair-mode options already contain both an old and replacement key, the upgrade retains both entries and does not overwrite either value; the replacement key stays authoritative and the stale key is left for manual cleanup.

### Version bump to 2.12.2
`ultimate-member.php` header, `readme.txt` stable tag, changelog entries. A package registers only when its version is <= UM_VERSION, so the bump is required for the upgrade to appear.

## Testing
**Test 1: migrate a site saved with the old names**
1. Install UM 2.10.0 or older, add the Country field to the registration and profile forms, register users with "Turkey" selected, optionally add a custom multiselect with the old names and a directory filtered by "Turkey"
2. Install this build over it and run Ultimate Member > Upgrade
3. Result: form dropdowns list the new names, profiles display the saved values again, the directory country filter matches the migrated users, and conditional rules referencing the old value keep working

Migration logic also verified with standalone checks covering exact matching, serialized multiselect values, the submitted copy, old/new key collisions, and unrelated callback fields. The repo has no unit test suite to extend.

Fresh installs are not affected: the package only shows as pending when `um_last_version_upgrade` is below 2.12.2.


ashubawork:
@faisalahammad 
I would not merge this as-is yet because the migration is too broad for a data upgrade routine.

Main concern: `um_country_name_replace_value2122()` recursively replaces matching strings anywhere in the field config. We should avoid rewriting the whole field array and target only the places that are meant to contain selectable country values: options, defaults, conditional logic values, member directory default filters, and relevant usermeta. Replacing arbitrary exact strings in the full field config can unexpectedly change labels, titles, placeholders, or other custom field settings.

A second concern is coverage: custom country-like fields that use the built-in countries list but do not have old country labels persisted in `_um_custom_fields` may not be added to `$fields_for_upgrade`, so their usermeta could remain unmigrated.

The direction is good, but for a persistent data migration we should make the replacement scope narrower and easier to reason about before merging.

faisalahammad:
Thanks for the review. Fixed both points in daed32d0f.

Scope: replacement now only touches options, default (on select/multiselect/radio/checkbox fields), conditional_value keys, and conditions tuples. Labels, titles, placeholders and help text are not touched anymore.

Coverage: fields with custom_dropdown_options_source set are now always added to the usermeta migration list, even when they have no static options snapshot to match against.

Directory filters and usermeta replacement stay the same, since you said those are fine targets.

Ready for another look.

ashubawork:
@faisalahammad 

Thanks, this version is much better. The migration scope is now much safer: it no longer walks the whole field config and avoids rewriting labels, titles, placeholders, and other display text. Targeting options/defaults, conditional values, directory filters, submitted data, and usermeta is the right direction.

I have one remaining concern before merging. Fields with `custom_dropdown_options_source` are now always added to the usermeta migration list, but a custom dropdown source is not necessarily a country field. It can be any callback-driven select/multiselect field. If such a field has a saved value exactly equal to `Turkey`, this upgrade would rename it to `Türkiye` even if the field is unrelated to countries.

Can we narrow this to fields that are known to use the built-in countries source/list, or add a more explicit opt-in/check for country-like callback fields? If we intentionally keep the broader callback-field behavior, I think the PR description should call out that trade-off.

One smaller edge case: when pair-mode options already contain both the old and new country keys, the current logic avoids overwriting the new key but leaves the old key in place. That is safe from a data-loss perspective, but it may leave the old selectable value available. Please document the intended behavior for that collision case or handle it explicitly.

Apart from that, the revised approach looks good.

faisalahammad:
CodeRabbit review triage: the single finding about restricting the options rewrite to "recognized country fields" is a duplicate of the already-handled human review point. The maintainer asked to narrow generic callback fields, which I applied. The remaining guard intentionally targets choice fields with stored old country values, exactly as the author's approved design intends; there is no field flag that identifies a country field beyond the predefined `country` metakey, and exact-only string replacement cannot rewrite unrelated labels/options. No code change was made for this CodeRabbit finding.

faisalahammad:
Fixed in 128e9a3c3.

Generic custom_dropdown_options_source fields are no longer added to the country usermeta migration list. Only fields with an old country value in their stored config are queued, while the predefined country field remains covered explicitly. Pair-mode collisions are documented and preserve both old and replacement keys without overwriting either value.

I also verified the migration helpers, PHP syntax, and manual upgrade flow.

PR #1873, opened 2026-08-17
fix(directory): use LIKE compare for text filters with custom usermeta table

### LINKED ISSUE #1719: Directory “More filters” not providing results as expected
**Expected behavior**
Activating “Custom usermeta table” should give same Directory results as the WP `usermeta` setting

**Describe the bug**
I created a text field meta_key = `text_free` and included the field in the Directory "Choose filter(s) meta to enable"
Added User Profile field values like `abc 123 qwe`, `123 456 789` etc as normal text examples.
Directory filter search with WP `usermeta` table with `123` found two Users as expected.
Switching to “Custom usermeta table” and the Directory filter search found no Users.

**SQL differences**
The WP usermeta case: (class-member-directory.php: 2913)
```
( mt2.meta_key = 'text_free'
  AND mt2.meta_value LIKE '%123%' )
  )
```
The “Custom usermeta table” case: (class-member-directory-meta.php: 1071)
```
AND umm1.um_key = 'text_free'
AND umm1.um_value = '123'
```
This second case expects the field value to only contain single words!
Searching words in sentences is only supported for WP `usermeta` table setting.

**UM Support Forum issue**
https://wordpress.org/support/topic/directory-more-filters-not-providing-results-as-expected/

faisalahammad:
## Summary
Member directory text filters used exact match when the "Custom usermeta table" option is enabled, while the WP usermeta table path used a LIKE match. A field value like `abc 123 qwe` was never found when filtering for `123`. This changes the custom table path to use LIKE for frontend filters with `esc_like()` escaping, keeping exact match for admin default filters, so both storage modes return the same results.

Fixes #1719

### includes/core/class-member-directory-meta.php (`handle_filter_query()`, text case)
**Before:**
```php
$compare = apply_filters( 'um_members_directory_filter_text', '=', $field );
$compare = esc_sql( $compare );
$value   = apply_filters( 'um_members_directory_filter_text_meta_value', $value, $field );

$this->where_clauses[] = $wpdb->prepare( "{$join_alias}.um_key = %s AND {$join_alias}.um_value {$compare} %s", $field, $value );
```
**After:**
```php
$compare = apply_filters( 'um_members_directory_filter_text', $is_default ? '=' : 'LIKE', $field );
$compare = esc_sql( $compare );
$value   = apply_filters( 'um_members_directory_filter_text_meta_value', $value, $field );

$where_value = $value;
if ( false !== stripos( $compare, 'LIKE' ) ) {
    $where_value = '%' . $wpdb->esc_like( $value ) . '%';
}

$this->where_clauses[] = $wpdb->prepare( "{$join_alias}.um_key = %s AND {$join_alias}.um_value {$compare} %s", $field, $where_value );
```
**Why:** The usermeta path (`class-member-directory.php`, `filters()` for frontend filters and `default_filters()` for admin defaults) already uses `LIKE` and `=` respectively. This mirrors those defaults per context. The wildcard wrapping plus `esc_like()` matches how `WP_Meta_Query` builds LIKE clauses and follows the same pattern as the existing select case in this file. The raw value is still stored in `custom_filters_in_query`.

## Testing
**Test 1: Text filter with multi-word values, custom usermeta table enabled**
1. Enable the custom usermeta table in Ultimate Member > Settings > Misc
2. Create a text field (meta key `text_free`), set two users' values to `abc 123 qwe` and `123 456 789`
3. Add the field to a member directory under "Choose filter(s) meta to enable"
4. Open the directory page on the frontend, open More filters, type `123` in the text filter, apply

Result: both users are returned. Before the fix no users were returned.

Regression checks done: admin default filters still use exact match, filter types other than text are untouched, values containing `%` or `_` are matched literally, and the usermeta code path is unchanged. PHPCS (WordPress-Extra) reports no new violations for the file.


ashubawork:
@nikitasinelnikov 
This looks like a valid fix for the issue #1719. The current custom usermeta table path uses an exact match for frontend text filters, while the regular `wp_usermeta` path uses `LIKE`, so the same directory filter can return different results depending on the storage mode.

The approach looks good:
- frontend text filters now default to `LIKE`;
- default/admin filters still use `=`;
- `esc_like()` correctly protects literal `%` and `_` characters;

nikitasinelnikov:
Manually merged into the `development/2.9.x` branch to make the release in version 2.13.1. Cannot merge this PR to 2.9.x for some reason.

Thanks for the fix @faisalahammad. and review @ashubawork

PR #1874, opened 2026-08-17
fix(directory): exclude filter values from hidden role users

### LINKED ISSUE #1717: Exclude values ​​in directory filters from hidden users

**Describe the bug**
Filter field in the member directory displays the value that is filled in by the user for whose role the display of profiles in the list of users in the directory is disabled.

**To Reproduce**

1. For the User Profile form - the Country field has been added
2. For Member Directories > General Options > User Roles to Display - the checkbox is checked (example: Subscriber role)
3. For Member Directories > Search Options > Choose filter(s) meta to enable - added Country field
4. User with a role (example role Contributor) fills in the Country field in the profile with a value (example Belgium)
5. Users with a role (example role Subscriber) don't have the Country field with a filled value (example Belgium)
6. Go to the site on the member directory page, select a filter for the Country field - see what value is displayed (example Belgium)

**Screenshots**
Image

faisalahammad:
## Summary
Select filters in the member directory collected their values from all users, including users whose roles are excluded from the directory via "User Roles to Display". Such values appeared in the filter dropdown but returned zero members when selected. The values query is now restricted to users with the directory's displayable roles.

Fixes #1717

### Member Directory (`includes/core/class-member-directory.php`)
**Before:**
```php
$values_array = $wpdb->get_col(
    $wpdb->prepare(
        "SELECT DISTINCT meta_value
        FROM $wpdb->usermeta
        WHERE meta_key = %s AND
              meta_value != ''",
        $attrs['metakey']
    )
);
```
**After:**
```php
$directory_roles = get_post_meta( absint( $directory_data['form_id'] ), '_um_roles', true );
if ( ! empty( $directory_roles ) && is_array( $directory_roles ) ) {
    $values_array = $this->get_filter_values_in_roles( $attrs['metakey'], $directory_roles );
} else {
    // original unrestricted query kept as-is
}
```
**Why:** When the directory limits displayed users by role, filter options should only contain values that at least one displayed user has. This mirrors how the same function already limits `role_select` options to `_um_roles`. With no roles selected the behavior is unchanged.

The new `get_filter_values_in_roles()` method joins the capabilities usermeta and matches the directory roles with `LIKE '%"role"%'` through `$wpdb->esc_like()`, the same serialized-role matching pattern already used in `includes/core/class-member-directory-meta.php`. Multisite is handled via `get_blog_prefix()`. Applies to the front-end directory, the admin directory builder, and the "Admin filtering" default filters, since all `show_filter()` callers pass the directory `form_id`.

## Testing
**Test 1: Filter values from excluded roles (issue repro)**
1. Add a Country field to the User Profile form.
2. Member Directories > General Options > User Roles to Display: check Subscriber only.
3. Member Directories > Search Options > Choose filter(s) meta to enable: add Country.
4. Set a Contributor user's Country to a unique value (Belgium); no Subscriber user has it.
5. Open the member directory page and open the Country filter dropdown.
Result: Belgium is not listed. Values set by Subscriber users are listed and filtering by them returns the expected members.

**Test 2: No role restriction (regression check)**
1. In the same directory, clear the User Roles to Display selection, save, reload the directory page.
Result: the Country filter lists values from all users again, including Belgium.

PHPCS was run against master for the changed file with no new violations. Manual test of the repro above passed on a packaged build of this branch.

### DIFF
--- a/includes/core/class-member-directory.php
+++ b/includes/core/class-member-directory.php
@@ -858,15 +858,20 @@ if ( ! class_exists( 'um\core\Member_Directory' ) ) {
 					global $wpdb;
 
 					if ( $attrs['metakey'] != 'role_select' ) {
-						$values_array = $wpdb->get_col(
-							$wpdb->prepare(
-								"SELECT DISTINCT meta_value
-								FROM $wpdb->usermeta
-								WHERE meta_key = %s AND
-									  meta_value != ''",
-								$attrs['metakey']
-							)
-						);
+						$directory_roles = get_post_meta( absint( $directory_data['form_id'] ), '_um_roles', true );
+						if ( ! empty( $directory_roles ) && is_array( $directory_roles ) ) {
+							$values_array = $this->get_filter_values_in_roles( $attrs['metakey'], $directory_roles );
+						} else {
+							$values_array = $wpdb->get_col(
+								$wpdb->prepare(
+									"SELECT DISTINCT meta_value
+									FROM $wpdb->usermeta
+									WHERE meta_key = %s AND
+										  meta_value != ''",
+									$attrs['metakey']
+								)
+							);
+						}
 					} else {
 						$users_roles = count_users();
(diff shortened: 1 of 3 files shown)


ashubawork:
@faisalahammad, we added another fix in the next version of the UM.

PR #1875, opened 2026-08-17
fix(login): validate required custom fields

### LINKED ISSUE #1664: Required checkbox field in Login form
**Expected behavior**
A checkbox field set as required in the Login form should send an error message to the User that field is required for Login if not being ticked by the User

**Describe the bug**
Validation of checkbox fields in login forms are missing.

**To Reproduce**
1. Add a checkbox field to Login form
2. Set field as required
3. Goto Login
4. Try to login without clicking the required checkbox
5. Login successful

**Expected behavior**
UM Form Builder settings must be validated for all fields regardless of Form type.

**UM Support Forum issue**
https://wordpress.org/support/topic/login-form-checkbox-required-field-not-working/

faisalahammad:
## Summary
A required field added to a Login form in the Form Builder was never validated on submit, so users could log in without filling it. The login branch of `um_submit_form_errors_hook()` only ran the username and password checks, while the field-by-field validation (`um_submit_form_errors_hook_`) only ran for register and profile forms.

This adds `um_submit_form_errors_hook_login_custom_fields()` at priority 20 on `um_submit_form_errors_hook_login`. It checks every required custom field on the login form and adds an inline error when one is missing, which blocks login through the existing error gate in `um_submit_form_login()`. Core credential fields keep their native validation, fields hidden by conditional logic are skipped, and fields the visitor cannot view are skipped via `um_can_view_field()`, matching how the registration validator behaves.

Scope: this change validates required custom fields on login forms only. It does not add the full set of Form Builder field-type validation rules used by registration and profile forms.

Fixes #1664

### includes/core/um-actions-login.php
**Before:**
```php
if ( 'login' === $mode ) {
    do_action( 'um_submit_form_errors_hook_login', $submitted_data, $form_data ); // username/password only
    do_action( 'um_submit_form_errors_hook_logincheck', $submitted_data, $form_data );
} else {
    // register/profile ...
    do_action( 'um_submit_form_errors_hook_', $submitted_data, $form_data ); // field validation, never for login
}
```
**After:**
```php
function um_submit_form_errors_hook_login_custom_fields( $submitted_data, $form_data ) {
    // checkbox, radio, multiselect: must be present and non-empty in the submitted data
    // other field types: must not be blank
}
add_action( 'um_submit_form_errors_hook_login', 'um_submit_form_errors_hook_login_custom_fields', 20, 2 );
```
**Why:** Required checkbox, radio and multiselect fields submit no key at all when left empty, so the check covers both the missing key and a crafted empty value (`field[]=`), while valid `'0'` values still pass. The callback also guards missing form metadata, applies the existing custom-field filter, and safely handles unexpected submitted array values.

### includes/core/um-actions-form.php
Hook docblock updated to list the new priority-20 callback. No code change.

## Testing
**Test 1: Required checkbox blocks login**
1. Add a checkbox field to the Default Login form, set it as required
2. Log out and go to the login page
3. Enter valid credentials, leave the checkbox unticked, submit
Result: login is blocked, inline error ` is required.` shows under the field

**Test 2: Ticking the checkbox allows login**
1. Tick the checkbox and submit with valid credentials
Result: login succeeds

**Test 3: Regression, default login form**
1. Log in with the standard username/password form, then with a wrong password
Result: works as before, usual incorrect password error

**Test 4: Regression, registration form**
1. Submit the Default Registration form with a required checkbox unticked
Result: same required error as before this change


ashubawork:
@faisalahammad 
The direction looks good and the reported bug is valid: login forms currently bypass the generic custom field validation path, so required custom fields can be submitted empty.

Before merging, I think we should tighten the implementation a bit:

1. Guard `custom_fields` before reading it from `$form_data`, to avoid introducing a new undefined array key warning.
2. Apply the existing `um_get_custom_field_array` filter inside the loop, matching the generic register/profile validation flow.
3. Avoid `array_filter( $value, 'strlen' )` for submitted array values, because unexpected nested/non-string values can trigger warnings on PHP 8+. A strict callback is safer.
4. Clarify the PR scope: this validates required custom fields on login forms, but it does not apply the full set of Form Builder validation rules.

With those adjustments, I think this is a good targeted fix for the required login custom field issue.

faisalahammad:
Thanks for the review. Applied all four points in commit 5fdad8d29:

1. Guard `$form_data['custom_fields']` before reading it.
2. Apply `um_get_custom_field_array` inside the per-field loop.
3. Replace `array_filter( $value, 'strlen' )` with a strict callback that keeps non-empty strings and numeric values.
4. Updated the PR description with a scope note: this only validates required custom fields on login forms, not the full Form Builder rule set.

Re-review when convenient.

ashubawork:
@faisalahammad 

Thanks, this is closer and the main issue is valid. The guard around `custom_fields`, the safer array handling, and the clarified scope all look good.

I still see one blocker before merging: the PR description says fields the visitor cannot view are skipped via `um_can_view_field()`, but the implementation does not call that filter or mirror the visibility/public-role checks used by the generic register/profile validator. It only skips fields with `visibility === 'view'`. That could make a required field block login even when the field is not actually rendered for the visitor.

Also, `um_get_custom_field_array` is applied after the condition checks, while the generic validator applies it before conditions. To keep login behavior aligned with register/profile custom field validation, I think the filter should run before checking conditions/required state.

faisalahammad:
@ashubawork Addressed in commit 3dab8f06b. The login validator now applies `um_get_custom_field_array` before visibility and condition checks, skips fields that fail `um_can_view_field()` while preserving the non-rendered visibility check, and keeps the existing required-field handling. Manual testing is confirmed.

faisalahammad (commented):
Implemented and verified the requested login custom-field validation updates.

- Applies `um_get_custom_field_array` before visibility, conditional, and required checks.
- Uses `um_can_view_field()` and retains the explicit non-rendered visibility check.
- Preserves required checkbox, radio, multiselect, and conditional-logic behavior.
- PHP syntax checks passed; CodeRabbit reported no findings.
- Manual testing was confirmed.

PR #1876, opened 2026-08-17
fix(fields): make gender field work on account page

### LINKED ISSUE #1698: Gender field at Account page
**Expected behavior**
Adding the UM predefined field gender to the Account page should display gender value from User Registration.

**Describe the bug**
Account page and Profile/Registration pages use different HTML setup for radio fields and the gender field is not prepared for this difference when displaying gender value from either Registration or Profile page.

**To Reproduce**
Add gender field to Account page
```
add_filter( 'um_account_tab_general_fields', 'um_account_tab_general_fields_gender', 10, 2 );
function um_account_tab_general_fields_gender( $args, $shortcode_args ) {
	$args = str_replace( ',single_user_password', ',gender,single_user_password', $args );
	return $args;
}
```
Try to edit gender in either Account or Profile page
Account page display and edit is not possible.

**Expected behavior**
Make the Account page gender field compatible with Profile/Registration DB format
```
add_filter( 'um_radio_field_options', 'um_radio_field_options_gender', 10, 2 );
function um_radio_field_options_gender( $options, $data ) {
    if ( $data['metakey'] == 'gender') {
        $options = array_combine( $options, $options );
    }
    return $options;
}
```
**Support Forum issue**
https://wordpress.org/support/topic/critical-custom-fields-not-displaying-selecting-correctly-on-account-page/

faisalahammad:
## Summary

The predefined `gender` radio field does not work on the Account page. Radio rendering uses the option array key as the submitted value in account context (`edit_field()` case `radio` in `includes/core/class-fields.php`), while Profile/Registration pages use the option value. The gender options were a plain numeric list, so the Account page rendered `value="0"` / `value="1"`, never showed the saved gender as selected, and stored `0`/`1` in usermeta.

This changes the gender options to key/value pairs where the key equals the value, so all three contexts produce and store the same label format.

Fixes #1698

### Gender field definition in `includes/core/class-builtin.php`

**Before:**
```php
'options'  => array(
	__( 'Male', 'ultimate-member' ),
	__( 'Female', 'ultimate-member' ),
),
```

**After:**
```php
'options'  => array(
	__( 'Male', 'ultimate-member' )   => __( 'Male', 'ultimate-member' ),
	__( 'Female', 'ultimate-member' ) => __( 'Female', 'ultimate-member' ),
),
```

**Why:** Account mode takes the array key as the radio value, so numeric keys can never match the label stored by Profile/Registration forms. Keys equal to values make both paths identical. This also matches how the other account page fields are defined (`hide_in_members`, `um_show_last_login`).

## Testing

**Test 1: Reproduce before the fix**
1. Add gender to the account General tab with the snippet below.
2. Register a user with a gender selected, then open the Account page.
3. No radio is selected; changing and saving writes `0`/`1` to the `gender` usermeta row.

**Test 2: Verify after the fix**
1. Open the Account page: the saved gender shows as selected.
2. Change gender on the Account page and save: usermeta stores the label (`Male`/`Female`), same format the Profile page stores.
3. Check Profile and Registration pages render and save exactly as before.
4. Member directory gender filter still matches users by label values.

Result: works as expected. Checked with PHPCS (no new violations on the changed lines) and tested manually on account, profile and registration flows.

Snippet to add gender to the account page (from the issue):

```php
add_filter( 'um_account_tab_general_fields', 'um_account_tab_general_fields_gender', 10, 2 );
function um_account_tab_general_fields_gender( $args, $shortcode_args ) {
	$args = str_replace( ',single_user_password', ',gender,single_user_password', $args );
	return $args;
}
```


ashubawork:
@nikitasinelnikov 
This looks like a valid minimal fix for issue #1698.

The root cause is that account-page radio rendering uses the option array key as the submitted value, while the predefined `gender` field currently has numeric option keys. That makes the Account page submit/store `0` or `1`, while Profile and Registration store the label values (`Male` / `Female`). Making the option keys match the option values keeps all contexts using the same stored format.

nikitasinelnikov:
Scheduled to 2.14.0 version due to DB upgrade script is needed. It's not a good idea to store the translated value in the DB and use it as an option key.

Better to use:

```
'options' => array(
   'male'    => __( 'Male', 'ultimate-member' ),
   'female' => __( 'Female', 'ultimate-member' ),
)
```

But an upgrade script is needed. @ashubawork, you can add this task to your to-do.

Best Regards!

PR #1879, opened 2026-08-24
Very slow SQL query (CU-86d44dzab)

yuriinalivaiko:
The SQL query execution may take minutes in `um\common\actions\Users::batch_check()`
This pull request contains an optimization of this query.


nikitasinelnikov:
Native WordPress query looks like:

SELECT DISTINCT SQL_CALC_FOUND_ROWS wp_users.ID FROM wp_users LEFT JOIN wp_usermeta ON ( wp_users.ID = wp_usermeta.user_id AND wp_usermeta.meta_key = '_um_registration_in_progress' ) LEFT JOIN wp_usermeta AS mt1 ON ( wp_users.ID = mt1.user_id ) LEFT JOIN wp_usermeta AS mt2 ON ( wp_users.ID = mt2.user_id AND mt2.meta_key = 'account_status' ) LEFT JOIN wp_usermeta AS mt3 ON ( wp_users.ID = mt3.user_id ) WHERE 1=1 AND ( ( wp_usermeta.user_id IS NULL OR ( mt1.meta_key = '_um_registration_in_progress' AND mt1.meta_value != '1' ) ) AND ( mt2.user_id IS NULL OR ( mt3.meta_key = 'account_status' AND mt3.meta_value = '' ) ) ) ORDER BY user_login ASC LIMIT 0, 50

We don't use ORDER BY, SQL_CALC_FOUND_ROWS and use only 2 LEFT JOIN instead of 4

PR #1887, opened 2026-09-08
Fix/wp config writable issue fallback

champsupertramp:
The initial API keys implemented to the wp-config.ph doesn't work for those with newly updated core. There should be a fallback to the wp_options to pull them and later opt-in to the wp-config.php with the prompt notice.


nikitasinelnikov:
Hi @champsupertramp 
I saw you ignored api keys checkinh during admin-ajax.php running.
Probably we also need to ignore during WP_CRON and REST_REQUEST?

champsupertramp:
Hi @nikitasinelnikov  please see the latest commit https://github.com/ultimatemember/ultimatemember/pull/1887/commits/a214f7ec0cfb371ed633c06665978ac78d60b45f

Thank you

